PR #8, opened 2022-12-22
CodeGen, Type Class Deriving, Proto Compatibility Types (+ TH "frontend")

gnumonik:
Update: 

This PR has moved substantially beyond the "experiment" phase and is now a monolithic PR for the code generator, typeclass deriver/constraint solver, proto compatibility types, and a bonus template haskell "frontend".


gnumonik:
@bladyjoker 

I think this might be ready for you to take a look at. Bear in mind that it's still very first-draft-ey and the details of everything can change. Having said that, I think this might be a good idea once refined. 

At the moment the `Gen` stuff is capable (at a very basic / primitive level) of generating Rust type definitions from patterns. It's a bit ugly, and there's a lot of stuff it doesn't support, but it works for basic structs and enums containing primitive types, and can translate (something like) the Source type you drafted a while ago into a pattern form. 

`Common.Types` contains the ADT for patterns (and a bunch of pattern synonyms) + some utils and translation functions. Start there. 

`Resolve.Rules` contains the ADTs to represent instances and classes, and `Resolve.Solver` contains the basic constraint solver. 

`Common.Match` contains the pattern matcher (technically it checks whether the second arg is a substitution instance of the first). This is used by both the codegen and the constraint solver. 

`Gen.Generator` contains the basic type (`Gen`) and functions for code generation. It's *extremely* simple. Like, it'll take you 5 minutes to understand it tops. 

`Gen.Rust` contains the primitive rust code generators. This is probably what you want to spend most of the time on b/c it'll give you an idea of whether you think my approach is worth using. You can try it out on the test types in `Common.SourceTy` by using `genFromDefAndPrint` from the Gen module. 

#### Notes (for both of us, I guess): 
 - It doesn't generate typeclasses (well traits I guess) yet but I'm pretty sure I can just bundle generators with instance rules and make it do that fairly easily. A `Map Instance Gen` and when you solve a constraint you run the generator for it. (This is how we can just derive every intermediate instance for every component type.)

 - We need to figure out imports. We should actually be capable of gathering all the necessary imports _before_ we start generating code, I think? But we don't seem to have any place for users to define the _package_ in which a foreign module  is located, so we can't actually generate any kind of language-specific package manager stuff until we do that. 

 - I need to generate some kind of pretty printer library's `Doc` type instead of `Text`. As much as I hate them, they're necessary here; it's a bit too difficult to make things look as pretty as they should with just `Text` output.

typeclass-experiments/src/Gen/Rust.hs
```diff
@@ -0,0 +1,161 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Gen.Rust where
```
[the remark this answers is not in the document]
bladyjoker: Yeah, I think mapping directly to Text/Doc is not a good idea, we need to make this very clean, especially when we start mapping to actual Rust builtins, which means we need to manage module/package information and all that stuff.

Concatenating string at this stage seems like a foot gun. But it doesn't really change much, we just parametrize the generator with a Rust DSL type.

typeclass-experiments/src/Gen/Rust.hs
```diff
@@ -0,0 +1,161 @@
...
+    , rustRef
+    , rustMaybe
+    , rustList
+    , rustBareType
+    , rustRecFields
+    , rustRecConstr
+    , rustTupleBody
+    , rustTupleConstr
+    , rustTypeNoFreeVars
+    ]
```
bladyjoker, line 161: This is quite confusing, I was going through this thinking it's all about typeclasses, and this is all about type definition mapping? Is that somehow covered or did I just misread?

bladyjoker (commented):
Let's review this interactively during a meeting, it's hard to make sense out of this just by reading through. I think I get bits and pieces I just don't know what's solving what.

typeclass-experiments/src/Common/Types.hs
```diff
@@ -0,0 +1,175 @@
...
+
+{- A simple ADT to represent types (i.e. the terms of our schema language).
+
+We can also use this to represent patterns by using PatVars in compound types.
+
+Note that this ADT allows us to represent nonsensical types. I wanted to prove to Drazen that I could
+do something without using a billion language extensions, and some nonsense is the price we pay for that :p
+-}
+
+data Pat
```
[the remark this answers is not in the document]
bladyjoker: Yeah I still don't understand why you're reformulating everything into Pat? It looks isomorphic.

What are you actually trying to do? Have a query on types or? Add some documentation and examples, because it doesn't click for me.

If you're trying to do querying on types you can definitely use https://hackage.haskell.org/package/unification-fd-0.11.2/docs/Control-Unification.html and represent our types with UTerm and Variable. You get the ability to 'pattern match' ie. do basic unification query on pairs of type terms. Is that at all what you're trying to do?

typeclass-experiments/src/Common/Types.hs
```diff
@@ -0,0 +1,175 @@
...
+toRec = RecP . foldr (:*) Nil
+
+toSum :: [Pat] -> Pat
+toSum = SumP . foldr (:*) Nil
+
+{- Utility functions. Turn a list of types into a product/record/sum type.
+-}
+defToPat :: Ty.TyDef -> Pat
+defToPat Ty.TyDef {..} = DecP (Name tyDefName) vars $ case tyDefBody of
+  Ty.Sum constrs -> toSum . NE.toList . fmap goConstr $ constrs
```
bladyjoker, line 141: This just maps TyDef to Pat? How are they different?

typeclass-experiments/src/Gen/Generator.hs
```diff
@@ -0,0 +1,68 @@
...
+        -- This lets us handle infinite applications of TyCons,
+        -- e.g. Maybe (Maybe (Maybe (Maybe IntP)))
+        let gen = Gen p $ \scope px -> f (S.insert g scope) px
+        modify $ S.insert gen
+        newRules <- get
+        lift $ runGen gen newRules pat
+      xs -> lift . Left $ MultipleMatches pat (map genPat xs)
+  where
+    matchRules :: Set Gen -> [Gen]
+    matchRules xs = S.toList $ S.filter (\gx -> matches (genPat gx) pat) xs
```
bladyjoker, line 63: I sense this is the key to this whole PR, but staring at it for 15minutes doesn't reveal what it does. Please formulate this clearly and provide input output examples and elaborate concisely on the purpose.

typeclass-experiments/src/Gen/Generator.hs
```diff
@@ -0,0 +1,68 @@
...
+  | MalformedPatList Pat
+  deriving (Show)
+
+type GenResult = Either GenError Text
+
+-- crude code generator
+data Gen = Gen
+  { genPat :: Pat
+  , runGen :: Set Gen -> Pat -> GenResult
+  }
```
bladyjoker, line 27: This could be called `TypeDefPrinter`? As in, this just maps type definitions to Text? Just trying to form a mental map of this.

As mentioned elsewhere, let's not map directly to Text as it's very unmaintainable and overly complex, instead formulate a concise Rust DSL so we can talk in terms of RustModule, RustTypeDeclaration, RustFunction, RustBultinXYZ, RustStatement etc. And afterwards we can have a super simple Text printing facility implemented.

typeclass-experiments/src/Gen/Generator.hs
```diff
@@ -0,0 +1,68 @@
...
+      [] -> lift . Left $ DeriveFail pat
+      [g@(Gen p f)] -> do
+        -- The trick is that we make sure every generator always has access to itself.
+        -- This lets us handle infinite applications of TyCons,
+        -- e.g. Maybe (Maybe (Maybe (Maybe IntP)))
+        let gen = Gen p $ \scope px -> f (S.insert g scope) px
+        modify $ S.insert gen
+        newRules <- get
+        lift $ runGen gen newRules pat
+      xs -> lift . Left $ MultipleMatches pat (map genPat xs)
```
bladyjoker, line 60: How can it `match` on multiple? How would that work?

typeclass-experiments/src/Gen/Rust.hs
```diff
@@ -0,0 +1,161 @@
...
+  Maybe t -> do
+    t' <- runGenerator scope t
+    pure $ "Option<" <> t' <> ">"
+  other -> matchFail (Maybe _x) other
+
+rustList :: Gen
+rustList = Gen (List _x) $ \scope pat -> case pat of
+  List t -> do
+    t' <- runGenerator scope t
+    pure $ "Vec<" <> t' <> ">"
```
bladyjoker, line 75: With Rust DSL this would be `pure $ RustBuiltinVector t'`

typeclass-experiments/src/Common/Types.hs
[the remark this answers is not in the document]
gnumonik: Because it *is* a pattern, and not a type. It can have holes for variables in places where you can't put a type variable. It becomes a type when it gets instantiated such that the only variables in the pattern are in places where type variables can legitimately live, or when there aren't any variables in the pattern.

typeclass-experiments/src/Common/Types.hs
```diff
@@ -0,0 +1,175 @@
...
+  | {- arity 0 primitives -}
+    IntP
+  | StringP
+  | BoolP
+  | {- arity 1 primitives -}
+    ListP
+  | MaybeP
+  | {- arity 2 primitives -}
+    MapP
+  | EitherP
```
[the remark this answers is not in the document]
gnumonik: This is actually compatible w/ not having special primitive types in the Source, for whatever that's worth. It's just an aesthetic thing.

typeclass-experiments/src/Common/Types.hs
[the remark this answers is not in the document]
gnumonik: When I say "it is a pattern", I mean it's analogous to [Pat in the TH AST](https://hackage.haskell.org/package/template-haskell-2.19.0.0/docs/Language-Haskell-TH-Syntax.html#t:Pat). Which is similar, but not identical to [Exp in the TH AST](https://hackage.haskell.org/package/template-haskell-2.19.0.0/docs/Language-Haskell-TH-Syntax.html#t:Exp).

typeclass-experiments/src/Common/Match.hs
```diff
@@ -0,0 +1,31 @@
...
+   with the argument type.
+
+   The first argument is the inner Pat from an instance head or Gen.
+
+   The second argument is the Pat representation of a type that we want to derive an instance / generate code for.
+
+   NOTE: Is not bidirectional! The first Pat has to be more general than the first
+         (more specifically: The second Pat should be a substitution instance of the first)
+-}
+matches :: Pat -> Pat -> Bool
```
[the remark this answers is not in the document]
gnumonik: I guess I could? `matches` should be faster than substitution if we only care about filtering things that match a pattern.

typeclass-experiments/src/Gen/Rust.hs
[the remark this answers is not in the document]
gnumonik: Well, this is kind of an attempt to get started on an eDSL that does that I suppose.

I think `Doc` is fine in the sense that: When we look at a schema and try to generate some code for it, we don't need anything outside of that schema to generate the textual representation of the code, so we *should* be able to generate the prettily-formatted `Doc` straightaway. 

But since, in the course of generating code for a schema, we might also need to generate an import statement (or whatever), Doc can't be the _only_ thing that the generator outputs. We need generator functions to return something that's (morally) `(Doc,[Import])`. We probably can't generate `Doc` representations of imports directly because we'll want to do things like combine imports from the same module into one import statement & eliminate redundant imports & so on. 

We will need a set of utility functions for each target language for formatting though. Like, we'll need a DSL (well an eDSL) for bracketing and indentation conventions particular to rust. 

The advantage of this approach going forward, btw, is that we can write little modular code generators vs doing everything in massive case expressions on the `Source` like PS-bridge does. We *could* just do that, but it would be really unreadable and hard to debug and maintain - I think that's one lesson we learned from ps-bridge. 

I think we _need_ something like this approach; once we start implementing dozens (hundreds?) of opaque types (e.g. for the ledger), it will become a nightmare to handle every special case in a monolithic function. 

Also, we should be able to specify a finite number of patterns such that if we have a corresponding generator for each then we know we have a complete generator for a target language.

typeclass-experiments/src/Resolve/Solver.hs
```diff
@@ -0,0 +1,159 @@
...
+getSubs (x :* xs) (x' :* xs') = getSubs x x' <> getSubs xs xs'
+getSubs (l := t) (l' := t') = getSubs l l' <> getSubs t t'
+getSubs (Map k v) (Map k' v') = getSubs k k' <> getSubs v v'
+getSubs (Either l r) (Either l' r') = getSubs l l' <> getSubs r r'
+getSubs (ProdP xs) (ProdP xs') = getSubs xs xs'
+getSubs (RecP xs) (RecP xs') = getSubs xs xs'
+getSubs (SumP xs) (SumP xs') = getSubs xs xs'
+getSubs (AppP t1 t2) (AppP t1' t2') = getSubs t1 t1' <> getSubs t2 t2'
+getSubs (RefP t) (RefP t') = getSubs t t'
+getSubs _ _ = []
```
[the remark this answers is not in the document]
gnumonik: I guess you can think of it like that, though I was just thinking of it as normal logical substitution (which is a kind of simple unification I suppose)

typeclass-experiments/src/Gen/Rust.hs
gnumonik: I guess I could have been clearer in the initial comment to this PR, so lemme try again. This library contains 3 related things: 

 - A representation of patterns for our `Source` type. (`Common.Types`)
 - A constraint solver that makes use of those patterns to solve typeclass constraints. (`Resolve._`)
 - A code generation framework that uses those patterns to allow for modular definitions of code generation functions. (`Gen._`)
 
Because we need patterns anyway to do anything useful with typeclass constraints, I thought it would be useful to explore their relevance to generating code for the types themselves. Modulo some awkwardness resulting from using `Text` instead of `Doc` I think it actually works quite nicely? 

The idea is: If we're going to match on patterns of our SOP-like representation to generate/derive typeclass instances, it makes a lot of sense to use the same machinery to generate the type definitions in a target language as well.

typeclass-experiments/src/Gen/Generator.hs
gnumonik: `ProdP (IntP :* Nil)` would match both `ProdP (VarP "x" :* Nil)` and `ProdP (VarP "x")`. It's an error b/c rules should be unambiguous. (We could have a notion of a more/less specific match but it would be superfluous)

typeclass-experiments/src/Gen/Rust.hs
gnumonik: What do we get out of reifying that into a data type? I feel like that just adds a whole lot of code for no real benefit. See above comment about having `Gen` return `(Doc,[Import])`. I think that approach would end up saving us a lot of typing and ultimately be just as maintainable. It just seems like another layer of indirection to me.

typeclass-experiments/src/Common/Types.hs
gnumonik: It has more holes and it's flat. E.g. You can put a `VarP "x"` _anywhere_ in a pattern (more or less), which gives us a first class representation of things like "This is a list of one element" (which is relevant when generating rust types b/c turning `data F = F Int` into `enum F { F(Int) }` is super unidiomatic) 

I mean generally it's the difference between a pattern and an expression? I don't think I can do justice to the whole point of patterns in a reply here. If Haskell were more advanced we could just use Haskell patterns, but they aren't first class so we can't pass them around, which we need to do for constraint solving at least.

gnumonik: If you really want to use that Control.Unify library I can give it a go but I think it would just make a lot of our code more confusing than it needs to be. 

`Pat` is NOT isomorphic to `Ty`. You can have something like `DecP (Name "Name") Nil (VarP "body")` and you _can't_ turn that into a `Ty` without instantiating the `VarP "body"` to some concrete TyBody. Maybe that helps explain why there are "more holes" and why that's useful.

typeclass-experiments/src/Common/Types.hs
```diff
@@ -0,0 +1,175 @@
...
+      Ty.Product pList -> toProd . NE.toList . fmap goPat $ pList
+
+    goPat :: Ty.SourceTy -> Pat
+    goPat = \case
+      Ty.TyVar t -> VarP t
+      Ty.TApp a b -> AppP (goPat a) (goPat b)
+      Ty.PrimT prim -> goPrim prim
+      Ty.TyRef ref -> case ref of
+        Ty.Local t -> RefP (Name t)
+        _ -> undefined -- dunno what to do here. Imports should be resolved by this stage?
```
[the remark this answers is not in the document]
gnumonik: Oh I misunderstood how foreign refs work I think. You're saying that foreign refs are refs to other *LambdaBuffs* modules and not to other modules in the target language, right? 

Isn't it a big problem that we don't allow users to import types from other modules in the target languages? That seems like a *much* bigger practical limitation than e.g. the lack of recursive types would be. 

I guess they could define them as opaque types... but if we let them *use* those types it would break our ability to derive typeclass instances unless the users submitted a PR implementing instance rules for them. 

As annoying as ps-bridge was, the fact that you could specify "translate Haskell type `X` into PS type `Y` from module `Z`" was borderline essential for our use of it, was it not?

typeclass-experiments/src/Gen/Generator.hs
gnumonik: It's a way of bundling modular code generation functions with a first class representation of the pattern that they generate code for. This _is a framework for writing a DSL like that_. Like, here, example: 

``` haskell 
rustList :: Gen
rustList = Gen (List _x) $ \scope pat -> case pat of
  List t -> do
    t' <- runGenerator scope t
    pure $ "Vec<" <> t' <> ">"
  other -> matchFail (List _x) other
```

`List _x` is just sugar for `AppP ListP (VarP "x")` or, if you really wanna do things with no builtins, `AppP (RefP "List") (VarP "x")`. What that means is: "I'll match with any [a], no matter what [a] is". That's the part that determines which pattern the generator is a generator for. 

The function part takes a `Set Gen` (`scope` for lack of a better name) and a `Pat`. 

*We* know that the `Pat` is a substitution instance of `List _x / AppP ListP (VarP "x")` but we still have to match on it to get ahold of the Haskell variable `t` (in the Haskell AST it's a `Language.Haskell.TH.Syntax VarP`) which corresponds to the variable in our pattern. 

(Given enough time I could write some template haskell that converts our Pat into a TH Pat, or use fancy type stuff to enforce the structural similarity of the `Pat` in the function  to the `Pat` arg, but anyway let's pretend I won't do that for now since you hate that stuff. The _annoying_ part this about this is that we know that the `other` branch is inaccessible here but without fanciness we can't tell Haskell that it is)

The only thing we can do with `t` is try to generate some code for it, since `rustList` only knows how to wrap an argument type with `Vec<_>`. So we do that.

gnumonik: This is _kind of_ like a combination of a parser and pretty printer. The `genPat :: Pat` field in `Gen` tells you what to match (the parser), and the function part tells you what to construct from the match (the pretty printer). It *could* be 

``` haskell 
-- e is an error type, a is the result type 
data Gen e a = Gen
  { genPat :: Pat
  , runGen :: Set (Gen e a) -> Pat -> Either e a
  }
```

In principle though I don't think there would be a reason to do `Gen GenError RustDSL` instead of `Gen GenError ([Import],Doc)` though.

typeclass-experiments/src/Gen/Generator.hs
gnumonik:
``` haskell
-- some generators to use for the example 
rustInt :: Gen
rustInt = primGen IntP "i64"

rustString :: Gen
rustString = primGen StringP "String"

rustBool :: Gen
rustBool = primGen BoolP "bool"

rustMaybe :: Gen
rustMaybe = Gen (Maybe _x) $ \scope pat -> case pat of
  Maybe t -> do
    t' <- runGenerator scope t
    pure $ "Option<" <> t' <> ">"
  other -> matchFail (Maybe _x) other

rustList :: Gen
rustList = Gen (List _x) $ \scope pat -> case pat of
  List t -> do
    t' <- runGenerator scope t
    pure $ "Vec<" <> t' <> ">"
  other -> matchFail (List _x) other

myGens = S.fromList [rustInt, rustBool, rustString, rustMaybe, rustList]

-- this is just a convenience function so we don't have to type evalStateT everywhere 
runGenerator :: Set Gen -> Pat -> Either GenError Text
runGenerator gens pat = evalStateT (generator pat) gens

-- convenience 
runMyGens :: Pat -> Either GenError Text 
runMyGens = runGenerator myGens 
```
```
-- ghci 
ghci> runMyGens (List IntP)
Right "Vec<i64>"

ghci> runMyGens (List (List IntP))
Right "Vec<Vec<i64>>"

ghci> runMyGens (List (List (Maybe IntP)))
Right "Vec<Vec<Option<i64>>>"

ghci> runMyGens (Maybe (List (List (Maybe IntP))))
Right "Option<Vec<Vec<Option<i64>>>>"
```

I think that + everything else I wrote in all my other responses should make it clear what the purpose is.

typeclass-experiments/src/Common/Types.hs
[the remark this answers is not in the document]
bladyjoker: That's what `Opaque` types are, and as you suggested, you have to implement type-def mappings and whatever `instance` was declared for it. That's essentially the same, but devoid from ps-bridge hackery.

This is very much a 'LambdaBuffers FFI' feature, any and all primitive/builtin/opaques are treated the same way; if you declare an opaque, you must manually implement codegen aspects of it, namely type defs and typeclass semantics.

We should aim at the following Opaques:
1. Platform agnostic standard library, perhaps even getting feature parity with Protobufs (int, bytes, composites like arrays/maps/sets)
2. Plutus standard library that declares the entirety of the `plutus-ledger-api` types as Opaques

As we implement these, we will hopefully converge to an 'easy to use' codegen framework that is also robust (testing).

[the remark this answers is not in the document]
bladyjoker: We do not attack the use case of `I already have an application running with its own types, now I want to use LB` and take the route like others in this space where any shared types must be declared in LB from the very start.

typeclass-experiments/src/Common/Types.hs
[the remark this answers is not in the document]
bladyjoker: Ok thanks! I suspected you had something principled like this in mind, let me dig through the links you sent out and go through the whole thing again.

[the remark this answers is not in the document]
bladyjoker: So in TH Pat just represents 'pattern expressions' afaict, as in

```haskell
\Just (x:xs) -> ...
\pppppppp -> eee
```

Pat is `p` right?

typeclass-experiments/src/Common/Types.hs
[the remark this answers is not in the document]
bladyjoker: I'm not sure really. I think you're asserting what goes into the standard library and are treating this these 'special' types in a special manner, and that treatment is then inherited by all codegen modules.

I want to reflect the fact that LB Opaque types are all treated the same, none should get a special treatment so deep in the codegen framework by getting their special constructor.

In short, I would like to see `Opaque` and string matching on all the 'supported' type names.

typeclass-experiments/src/Common/Match.hs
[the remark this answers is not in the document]
bladyjoker: I think you're right yes. However, it's unclear to me atm whether this 'syntactical' equality is ignoring the variable names or not.

Basically, the difference between equals and equiv in https://hackage.haskell.org/package/unification-fd-0.10.0.1/docs/Control-Unification.html

typeclass-experiments/src/Gen/Rust.hs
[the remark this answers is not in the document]
bladyjoker: Not sure we're talking about the same thing. Consider a Rust DSL:

```
module LambdaBuffers.Rust where

data Module = Module {
  tyDefs :: TyDef,
  traitDefs :: TraitDef
}

data TyDef = TyDef {
  tyName :: String,
  ...
  tyBody :: tyBody
}

data Ty = ... | ... | TyRef TyRef

data Val = ValAbs .. | ValApp Val Val | ValRef ...
```

I would like to formulate what Rust IS from our point of view, to a very coarse level of detail, such that we map from LB to that ADT first. This is very informative, separates concerns such that it makes explicit the following to tasks:
1. How are LB constructs mapped to Rust constructs,
2. How are Rust constructs serialized to Text.

I feel this to be crucial to sustainability, opens up a neat testing opportunity, makes clear the different tasks the codegen does.

Bare in mind that Pat is still used to construct a `Rust.Module` from `LB.Module`, but you're not concatenating Text and messing with formatting at this point. And also note that you can perform 'fresh' variable and function naming at this stage, such that when you get to transforming `Rust.Module` to `Text` you don't have to do any heavy lifting.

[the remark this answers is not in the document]
bladyjoker: >The advantage of this approach going forward, btw, is that we can write little modular code generators vs doing everything in massive case expressions on the Source like PS-bridge does. We could just do that, but it would be really unreadable and hard to debug and maintain - I think that's one lesson we learned from ps-bridge.

I'm completely sold on that.

typeclass-experiments/src/Gen/Rust.hs
bladyjoker: That sounds great

typeclass-experiments/src/Gen/Generator.hs
bladyjoker: I'm liking this! I do still think we'd benefit from Rust DSL and adopting that approach throughout

typeclass-experiments/src/Gen/Rust.hs
bladyjoker: Clarity, testing, robustness, separation of concern. I think it's useful as looking at the Rust DSL module it would 'teach me' all I need to know what Rust is to a necessary level of detail.

I'm also thinking how easy it would be to make property or even model testing of Lb -> Rust mappings, and also we don't have to rely on tests based on Text matching as much.

That's my argument for this indirection, and generally I'm wary of formatting text anywhere close to 'core' logic, and it's generally recommended to delegate 'text rendering' to a very small and scoped module.

typeclass-experiments/src/Common/Match.hs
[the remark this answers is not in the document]
bladyjoker: I think I'm again misreading things, this could be an `Eq` on Pat if it wasn't for non-bidirectinality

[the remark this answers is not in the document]
bladyjoker: I think you should rename this to `subsumes`?

typeclass-experiments/src/Common/SourceTy.hs
```diff
@@ -0,0 +1,166 @@
...
+  | TP1 Prim1
+  | TP2 Prim2
+
+data SourceTy = TyVar VarName | TApp SourceTy SourceTy | TyRef TypeRef | PrimT TyPrim
+
+data TyDef = TyDef
+  { tyDefName :: TyConName
+  , tyDefKind :: TKind
+  , tyDefBody :: TyBody
+  }
```
bladyjoker, line 61: I'm curious why introduce the generator machinery and how does it differ to a vanilla approach by using just-a-function. I see you machinery as a 'rule' system that 'matches' on certain patterns and performs a text conversion, but I don't necessarily see what kind of modularity it brings compared to the following toy approach:

The only 'extension' point is the `opaques` input to `runCodegen` functions where you can add additional opaques as you extend the capabilities.

```haskell
-- These are all running under MonadFail m and MonadState m
-- Imagine the state containing PursModuleName, PursImports and PursExports for the current LB module being codegened

data PursCodegenState = S {
  opaques :: Map TyName (PursModuleName, PursTyName),
  imports :: Map TyName (PursModuleName, PursTyName),
  exports :: Map TyName [Constructor]
}
runCodegen :: Map TyName (PursModuleName, PursTyName) -> CompilerInput -> m Text
runCodegen opaques inp = evalState (S opaques [] [])

toPursModule (Module modName tyDefs _) = do
  setModName modName
  traverse toPursTypeDef tyDefs

toPursTypeDef (TyDef n k b) = do
    n' <- toPursTyName n
    k' <- toPursKind k
    b' <- toPursBody b
    addTyDefExport $ TyDef n k b -- module ModuleName (Foo(..)) etc.
    return $ Text.format "data %s %s = %s" (n', k', b')

toPursTy (TyVar v) = toPurs v
toPursTy (TyApp tyFun tyArgs) = do
    tyFun' <- toPursTy tyFun
    tyArgs' <- toPursTy tyArgs
    return $ Text.format "%s %s" (tyFun', tyArgs')

toPursTy (TyRef r) = do -- Here's where we need an extension option
  ty <- findTyByRef r
  if isOpaque ty then do
    importOpaque ty # this is just consulting a Map TyName (PursModuleName, PursTyName) or raises an error if not found, then adds it to the state PursImports
    
toPursBody (Sum ...) = ...
toPursBody Opaque = ...
```

bladyjoker: This is a toy as the actual tricky business comes with generating instances and functions, which will likely change this in some ways. But for type def codegen, this seems pretty sufficient.

NOTE: This is a brain dump, I didn't compile it or anything, just a scrabble

bladyjoker (commented):
I would like to discuss 2 points:

1. How does the approach basic functions in a some state monad fall short compared to your approach,
2. Benefits and shortcomings of DSLs for formulating statements in target languages.

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
```diff
@@ -0,0 +1,562 @@
...
+data TyInner
+  = TyVarI TyVar
+  | TyAppI TyApp
+  | TyRefI TyRef
+  deriving (Show, Eq, Ord, Generic)
+
+data Ty = Ty {
+  ty :: TyInner,
+  sourceInfo :: SourceInfo
+} deriving (Show, Eq, Ord, Generic)
```
bladyjoker, line 74: I think this is awkward. Why not make a Functor out of Ty like I did in the Frontend? https://github.com/mlabs-haskell/lambda-buffers/blob/main/lambda-buffers-frontend/src/LambdaBuffers/Frontend/Syntax.hs#L26

```
data Ty info
  = TyVar (VarName info) info
  | TyApp (Ty info) [Ty info] info
  | TyRef' (TyRef info) info
  deriving stock (Show, Eq, Ord, Functor, Foldable, Traversable)
```

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
```diff
@@ -0,0 +1,562 @@
...
+  | EmptySumBody SourceInfo
+  | EmptyName SourceInfo
+  | NegativeArity ArgName SourceInfo
+  | OneOfNotSet Text SourceInfo -- catchall
+  deriving (Show, Eq, Ord, Generic)
+
+class IsMessage (proto :: Type) (good :: Type) where
+  fromProto :: proto -> Either FromProtoErr good
+
+  toProto   :: good -> proto
```
bladyjoker, line 196: Do we really need a typeclass for this?

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
```diff
@@ -0,0 +1,562 @@
+{-# LANGUAGE KindSignatures, DuplicateRecordFields #-}
+{-# LANGUAGE DeriveGeneric, TypeSynonymInstances #-}
+{-# LANGUAGE DataKinds, OverloadedLabels #-}
+{-# OPTIONS_GHC -Wno-missing-kind-signatures #-}
+{-# OPTIONS_GHC -Wno-missing-deriving-strategies #-}
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
```
bladyjoker, line 6: Clean this up.

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
```diff
@@ -0,0 +1,562 @@
+{-# LANGUAGE KindSignatures, DuplicateRecordFields #-}
+{-# LANGUAGE DeriveGeneric, TypeSynonymInstances #-}
+{-# LANGUAGE DataKinds, OverloadedLabels #-}
```
bladyjoker, line 3: Why not add this to Cabal file

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
```diff
@@ -0,0 +1,562 @@
...
+
+data SourcePosition = SourcePosition {
+  column :: Int,
+  row    :: Int
+} deriving (Show , Eq, Ord, Generic)
+
+-- for our purposes we really only need a single name type
+data LBName
+  = LBName {name :: Text, sourceInfo :: SourceInfo}
+     deriving (Show, Eq, Ord, Generic)
```
bladyjoker, line 46: Wdym?

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
```diff
@@ -0,0 +1,562 @@
...
+
+type TyName     = LBName
+type ConstrName = LBName
+type ModuleName = LBName
+type ArgName    = LBName
+type VarName    = LBName
+type FieldName  = LBName
+type ClassName  = LBName
```
bladyjoker, line 54: Why not make this into proper types, easy to make a mistake where one name is used where other is expected.

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
bladyjoker: The different names are the result of different parsing/validation functions. The fact that they share the underlying Text is not making them the 'same' thing.

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
```diff
@@ -0,0 +1,562 @@
...
+
+-- NOTE: Theoretically this could overflow when converting back to Int32, but if you have an arity > 2147483647
+--       then you deserve the resulting errors
+type Kind = Natural
```
bladyjoker, line 58: https://github.com/mlabs-haskell/lambda-buffers/blob/main/lambda-buffers-proto/compiler.proto#L237

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
```diff
@@ -0,0 +1,562 @@
...
+  | NoInstanceHead ClassName SourceInfo
+  | NoConstraintArgs ClassName SourceInfo
+  | MultipleConstraintArgs ClassName [Ty] SourceInfo
+  | NoClassArgs ClassName SourceInfo
+  | MultipleClassArgs ClassName SourceInfo
+  | NoTyAppArgs SourceInfo
+  | EmptyRecordBody SourceInfo -- Ideally we should catch & rethrow this and the next one when we have access to the tyname
+  | EmptySumBody SourceInfo
+  | EmptyName SourceInfo
+  | NegativeArity ArgName SourceInfo
```
bladyjoker, line 189: We're ditching the arity encoding https://github.com/mlabs-haskell/lambda-buffers/blob/main/lambda-buffers-proto/compiler.proto#L237

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
```diff
@@ -0,0 +1,562 @@
...
+  typeDefs   :: [TyDef],
+  classDefs  :: [ClassDef],
+  instances  :: [InstanceClause],
+  sourceInfo :: SourceInfo
+} deriving (Show, Eq, Ord, Generic)
+
+newtype CompilerInput = CompilerInput {modules :: [Module]}
+  deriving (Show, Eq, Ord, Generic)
+
+data FromProtoErr
```
bladyjoker, line 178: This is cool, we should integrate this into compiler.proto

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
```diff
@@ -0,0 +1,562 @@
+{-# LANGUAGE KindSignatures, DuplicateRecordFields #-}
+{-# LANGUAGE DeriveGeneric, TypeSynonymInstances #-}
+{-# LANGUAGE DataKinds, OverloadedLabels #-}
+{-# OPTIONS_GHC -Wno-missing-kind-signatures #-}
+{-# OPTIONS_GHC -Wno-missing-deriving-strategies #-}
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
+module LambdaBuffers.ProtoCompat.ProtoCompat where
```
bladyjoker, line 7: Explicit exports please, this is not allowed.

bladyjoker (commented):
Wrt to proto-compat:

1. Why put that in a separate package? This is Compiler internal right? We should merge this with the compiler lib and expose the module if anyone needs to use it.
2. Why use OverloadedLabels and DuplicateRecordFields. Why not just makeLens on all the types?
3. Can we please parametrize these types, they will at the very beginning be associated with SourceInfo, but as the processing progresses, kind information will be attached and whatever other annotation might come up.

lambda-buffers-codegen/src/LambdaBuffers/Common/Types.hs
```diff
@@ -0,0 +1,158 @@
...
+toSum = SumP . foldr (:*) Nil
+
+{- Utility functions. Turn a list of types into a product/record/sum type.
+-}
+
+
+-- misc utilities, break out into a Utils module if this grows too much
+
+-- whyyyyyy isn't this in prelude
+for :: [a] -> (a -> b) -> [b]
```
cstml, line 152: It is
 
```
[for :: (Traversable t, Applicative f) => t a -> (a -> f b) -> f (t b)](https://hackage.haskell.org/package/base/docs/Data-Traversable.html#v:for) 
```
Use `Identity` as the Applicative ?

lambda-buffers-codegen/src/LambdaBuffers/Common/Types.hs
```diff
@@ -0,0 +1,158 @@
...
+for = flip map
+
+safeHead :: [a] -> Maybe a
```
cstml, line 155: listToMaybe :: [a] -> [Maybe](https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-Maybe.html#t:Maybe) a

lambda-buffers-proto-compat/lambda-buffers-proto-compat.cabal
```diff
@@ -0,0 +1,102 @@
+cabal-version: 3.0
```
cstml, line 1: Moved file into PR #23.

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
```diff
@@ -0,0 +1,562 @@
+{-# LANGUAGE KindSignatures, DuplicateRecordFields #-}
```
cstml, line 1: Moved file into PR https://github.com/mlabs-haskell/lambda-buffers/pull/23.

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
gnumonik: Well it's either that or we have a billion functions & try to remember the name of each one. Typeclass is simpler.

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
gnumonik: Updated this in #24

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
gnumonik: Er, sorry about that. Everything is cleaned up in #24.

lambda-buffers-codegen/src/LambdaBuffers/Common/Types.hs
gnumonik: Oh, neat, thanks!

lambda-buffers-proto-compat/src/LambdaBuffers/ProtoCompat/ProtoCompat.hs
gnumonik: > Wrt to proto-compat:
> 
>     1. Why put that in a separate package? This is Compiler internal right? We should merge this with the compiler lib and expose the module if anyone needs to use it.
> 
>     2. Why use OverloadedLabels and DuplicateRecordFields. Why not just makeLens on all the types?
> 
>     3. Can we please parametrize these types, they will at the very beginning be associated with SourceInfo, but as the processing progresses, kind information will be attached and whatever other annotation might come up.

We discussed 1 and 3 at the meeting but w/r/t 2: Maybe GHC finally got around to fixing this, but as I recall `makeLens` will make things compile a lot slower than Generic Lens will. 

Also, if there's a way to do something without template haskell then _ceteris paribus_ it's a better way to do things. (IMO, maybe you disagree)

gnumonik:
Almost all of this discussion is outdated and significant portions of this have been split out. Closing in favor of #23 / #24 / #8 to refocus discussion and attention.

PR #26, opened 2023-01-25
KC: Integrate Common Validated Proto Types

cstml:
- swap the existing interim types for the proto validated ones in the compiler 
- improve data structures - i.e. use `Map` instead of lists, `Text` etc.
- break up the inference into smaller modules.
- add tests.
- add some missing _low hanging_ class instances.


bladyjoker (approved):
Good stuff, please take a look at and integrate KC into the CLI https://github.com/mlabs-haskell/lambda-buffers/blob/389570110c645341dda0760dad36686f2e2b56fe/lambda-buffers-compiler/app/LambdaBuffers/Compiler/Cli/Compile.hs#L27

PR #27, opened 2023-01-26
Compiler/superclass cycles

gnumonik:
Added check for superclass cycles in the compiler. It's not integrated with the other compiler components yet (not sure if the kindchecker is in a finished state / how we want to tie all of the checks together)


bladyjoker:
@gnumonik let's close this one off and work in #29

gnumonik:
Git is a mess here, closing in favor of #29

PR #31, opened 2023-01-30
Compiler: Class Kind Check

cstml:
Class Kind Check for well declared instances.


cstml:
Work is paused on the above - until further notice.

PR #32, opened 2023-01-30
Kind Checker: CLI Integration

cstml:
- hook up the Kind Checker to the CLI too.
- propose error types.
- propose Compiler output format.
- tidy up some of the things used in the test into their own modules.

### DIFF
--- a/lambda-buffers-compiler/app/LambdaBuffers/Compiler/Cli/Compile.hs
+++ b/lambda-buffers-compiler/app/LambdaBuffers/Compiler/Cli/Compile.hs
@@ -8,13 +8,12 @@ import Data.ProtoLens.TextFormat qualified as PbText
 import Data.Text.Lazy qualified as Text
 import Data.Text.Lazy.IO qualified as Text
 import LambdaBuffers.Compiler.KindCheck (check)
-import LambdaBuffers.Compiler.KindCheck.Context (getAllContext)
 import LambdaBuffers.Compiler.ProtoCompat (
   FromProtoErr (NamingError, ProtoError),
   IsMessage (fromProto, toProto),
  )
 import LambdaBuffers.Compiler.ProtoCompat.Types qualified as ProtoCompat
-import Proto.Compiler as ProtoLib (CompilerInput, CompilerResult)
+import Proto.Compiler as ProtoLib (CompilerInput, CompilerOutput)
 import System.FilePath.Lens (extension)
 
 data CompileOpts = CompileOpts
@@ -37,7 +36,7 @@ compile opts = do
       ProtoError pe -> print $ "Encountered a proto error " <> show pe
     Right compIn' -> do
       print @String "Successfully processed the CompilerInput"
-      let result = either (RCompilerFailure . KCErr) (RCompilerOutput . CompilerOutput . fmap kindConvert . getAllContext) $ check compIn'
+      let result = check compIn'
       writeCompilerOutput (opts ^. output) (toProto result)
   return ()
 
@@ -53,7 +52,7 @@ readCompilerInput fp = do
       return $ PbText.readMessageOrDie content
     _ -> error $ "Unknown CompilerInput format " <> ext
 
-writeCompilerOutput :: FilePath -> ProtoLib.CompilerResult -> IO ()
+writeCompilerOutput :: FilePath -> ProtoLib.CompilerOutput -> IO ()
 writeCompilerOutput fp cr = do
   let ext = fp ^. extension
   case ext of
--- a/lambda-buffers-compiler/app/LambdaBuffers/Compiler/Cli/Compile.hs
+++ b/lambda-buffers-compiler/app/LambdaBuffers/Compiler/Cli/Compile.hs
@@ -7,9 +7,18 @@ import Data.ProtoLens qualified as Pb
 import Data.ProtoLens.TextFormat qualified as PbText
 import Data.Text.Lazy qualified as Text
 import Data.Text.Lazy.IO qualified as Text
-import LambdaBuffers.Compiler.ProtoCompat (FromProtoErr (NamingError, ProtoError), IsMessage (fromProto, toProto))
+import LambdaBuffers.Compiler.KindCheck (check)
+import LambdaBuffers.Compiler.KindCheck.Context (getAllContext)
+import LambdaBuffers.Compiler.ProtoCompat (
+  CompilerFailure (KCErr),
+  CompilerOutput (CompilerOutput),
+  CompilerResult (RCompilerFailure, RCompilerOutput),
+  FromProtoErr (NamingError, ProtoError),
+  IsMessage (fromProto, toProto),
+  kindConvert,
+ )
 import LambdaBuffers.Compiler.ProtoCompat.Types qualified as ProtoCompat
-import Proto.Compiler (CompilerInput)
+import Proto.Compiler as ProtoLib (CompilerInput, CompilerResult)
 import System.FilePath.Lens (extension)
 
 data CompileOpts = CompileOpts
@@ -19,6 +28,8 @@ data CompileOpts = CompileOpts
   deriving stock (Eq, Show)
 
 makeLenses ''CompileOpts
+
+-- NOTE(cstml) - let's use Katip instead of print.
 
 -- | Compile LambdaBuffers modules
 compile :: CompileOpts -> IO ()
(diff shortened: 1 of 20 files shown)


lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/Types.hs
```diff
@@ -217,6 +219,35 @@ data Module = Module
...
+  | UnificationErr Text
+  | RecursiveSubstitutionErr Text
+  deriving stock (Show, Eq, Ord, Generic)
+
+instance Exception InferenceErr
+
+data KindCheckErr
+  = InconsistentTypeErr TyDef
+  | InferenceFailure TyDef InferenceErr
+  deriving stock (Show, Eq, Ord, Generic)
```
bladyjoker, line 234: You're not attaching any SourceInfos. The whole point of them was so I can report back the exact thing that needs fixing. Please find a way to include SourceInfo.

lambda-buffers-compiler/app/LambdaBuffers/Compiler/Cli/Compile.hs
```diff
@@ -30,8 +41,8 @@ compile opts = do
       ProtoError pe -> print $ "Encountered a proto error " <> show pe
     Right compIn' -> do
       print @String "Successfully processed the CompilerInput"
-      writeCompilerOutput (opts ^. output) (toProto compIn')
-
+      let result = either (RCompilerFailure . KCErr) (RCompilerOutput . CompilerOutput . fmap kindConvert . getAllContext) $ check compIn'
```
bladyjoker, line 44: This is hard to read. What I want the CLI code to look like is to call `KindCheck.check compIn`, all the context building should happen in that function in the KindCheck module.

lambda-buffers-proto/compiler.proto
```diff
@@ -491,3 +489,107 @@ message CompilerInput {
...
+/* Inference Failure 
+ */
+
+message InferenceErr {
+
+  // Error referring to an unbound term. This usually means that the term was
+  // not defined.
+  message UnboundTermErr {
+    string msg = 1; 
+  }
```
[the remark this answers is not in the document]
bladyjoker: I assume this should be unpacked into,

```
message MissingTyRef {
  TyRef ty_ref = 1;
}

message MissingTyVar {
  TyRef ty_var = 1;
}
```

I understand that internally, your looking up a Term. But we don't speak in terms of Terms in the API, but Ty.

lambda-buffers-proto/compiler.proto
```diff
@@ -491,3 +489,107 @@ message CompilerInput {
...
+
+  // This should never happen.
+  message ImpossibleErr {
+    string msg = 1;
+  }
+
+  // Unification has failed - type is incorrectly defined.
+  message UnificationErr {
+    string msg = 1; 
+  }
```
bladyjoker, line 522: I don't know what to do with this error.

lambda-buffers-proto/compiler.proto
```diff
@@ -491,3 +489,107 @@ message CompilerInput {
...
+
+  // Infinite recursion due to a substitution replacing a term with a term
+  // containing itself. Should never type check.
+  message RecursiveSubstitutionErr {
+    string msg = 1;
+  }
```
bladyjoker, line 528: When does this happen?

lambda-buffers-proto/compiler.proto
```diff
@@ -491,3 +489,107 @@ message CompilerInput {
...
+
+/* Compilation Result 
+
+   Compiler Output is either an Error or a Compiler Output.
+ */
+message CompilerResult {
+  oneof compiler_result {
+    CompilerFailure failed_compilation = 1;
+    CompilerOutput compilation_result = 2;
+  }
```
bladyjoker, line 593: This naming is weird.

You can't have Result and then Output underneath it and then the CompilerOutput called compilation_result. Please clean this up. Imo

CompilerOutput = Either ComplierError Compiled
or
CompilerOutput = Either CompilerError CompilerResult

cstml: Ok - nested that result with the thinking that I'll leave the result extendable. I'll amend.

lambda-buffers-proto/compiler.proto
cstml: This happens when the inference creates a substitution where the substituted term contains the variable its being substituted with. I.e.:

Substitute `a` with `a -> a`. Any such type would be an incorrect type.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/Types.hs
cstml: Ok will think how to do it meaningfully.

lambda-buffers-proto/compiler.proto
```diff
@@ -491,3 +489,107 @@ message CompilerInput {
...
+}
+
+/*
+  Compiler Output
+
+  A successful Compilation Output.
+ */
+message CompilerOutput {
+  // A map of checked type definitions.
+  repeated TypeDefinition type_defs = 1;
```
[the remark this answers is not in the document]
cstml: The problem with the `TyDef` as it stands is that it holds too little information for it to be relevant (i.e. the module info is one level above - which is what would be relevant). Maybe there is a way to include it in the `Module`. Otherwise I thought this would be the nicest way to get the info.

lambda-buffers-compiler/test/Test/Samples/Proto/Helpers.hs
```diff
@@ -0,0 +1,93 @@
+module Test.Samples.Proto.Helpers (
```
bladyjoker, line 1: Let's use Utils or Aux?

lambda-buffers-compiler/lambda-buffers-compiler.cabal
```diff
@@ -114,6 +118,7 @@ library
 
   hs-source-dirs:  src
 
+-- note(cstml): should we name this something shorter? lb-cli?
 executable lambda-buffers-compiler-cli
```
bladyjoker, line 122: Yes even `lbc`?

lambda-buffers-compiler/src/LambdaBuffers/Compiler/KindCheck/Context.hs
```diff
@@ -14,9 +14,12 @@ import Prettyprinter (
   (<+>),
  )
 
+-- import LambdaBuffers.Compiler.ProtoCompat (TyRef)
+-- type Variable = TyRef
```
bladyjoker, line 18:
```suggestion
```

lambda-buffers-compiler/src/LambdaBuffers/Compiler/KindCheck/Variable.hs
```diff
@@ -1,6 +1,18 @@
-module LambdaBuffers.Compiler.KindCheck.Variable (Atom, Var) where
+module LambdaBuffers.Compiler.KindCheck.Variable (Variable (LocalRef, ForeignRef), Atom) where
 
 import Data.Text (Text)
+import Prettyprinter (Pretty (pretty), concatWith)
 
+-- import LambdaBuffers.Compiler.ProtoCompat (TyRef)
```
bladyjoker, line 6:
```suggestion
```

lambda-buffers-compiler/src/LambdaBuffers/Compiler/KindCheck/Variable.hs
```diff
@@ -1,6 +1,18 @@
...
 type Atom = Text
-type Var = Text
+
+-- type Variable = TyRef
```
bladyjoker, line 9:
```suggestion
```

lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/Types.hs
```diff
@@ -1,57 +1,60 @@
 {-# LANGUAGE DuplicateRecordFields #-}
+{-# OPTIONS_GHC -Wno-missing-import-lists #-}
```
bladyjoker, line 2: Remove if possible

lambda-buffers-proto/compiler.proto
```diff
@@ -491,3 +489,85 @@ message CompilerInput {
...
+
+  // This should never happen.
+  message ImpossibleError {
+    string msg = 1;
+  }
+
+  oneof internal_error {
+    ImpossibleError impossible_error = 1;
+  }
+}
```
bladyjoker, line 504: Can we have one InternalError and that's it?

bladyjoker (approved):
Good stuff

lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/Types.hs
cstml:
```suggestion
```

PR #34, opened 2023-02-01
Compiler: Arbitrary Instances + Property Test

cstml:
- Add Arbitrary instances to the `ProtoCompat` types.
- Add 1 Property test to the kind checker.


lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/Types.hs
```diff
@@ -44,142 +45,153 @@ module LambdaBuffers.Compiler.ProtoCompat.Types (
...
+      fn n = Kind <$> resize n arbitrary
+
+data KindType = KindRef KindRefType | KindArrow Kind Kind
   deriving stock (Show, Eq, Ord, Generic)
 
-data KindRefType
-  = KUnspecified
-  | KType
+instance Arbitrary KindType where
+  arbitrary = sized fn
```
bladyjoker, line 110: Why the indirection with `fn`?

lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/Types.hs
```diff
@@ -44,142 +45,153 @@ module LambdaBuffers.Compiler.ProtoCompat.Types (
...
+    where
+      fn :: (Num a, Ord a) => a -> Gen Ty
+      fn n
+        | n <= 0 = TyRefI <$> arbitrary
+        | otherwise =
+            oneof
+              [ TyVarI <$> arbitrary
+              , TyAppI <$> arbitrary
+              , TyRefI <$> arbitrary
+              ]
```
bladyjoker, line 138: How do you feel about keeping instances close to their type definitions. Imo it bloats a module that should be a pure type lib that we maintain as a highly readable place. I'd rather have a separate module where the Arbitrary instances are orphaned.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/Types.hs
```diff
@@ -228,14 +286,34 @@ data KindCheckError
...
 
 data CompilerResult = CompilerResult
   deriving stock (Show, Eq, Ord, Generic)
+  deriving (Arbitrary) via GenericArbitrary CompilerResult
 
 type CompilerOutput = Either CompilerError CompilerResult
+
+-- nonEmptyArbList :: forall a. Arbitrary a => Gen [a]
+-- nonEmptyArbList = getNonEmpty <$> arbitrary @(NonEmptyList a)
+
```
bladyjoker, line 306:
```suggestion
```

lambda-buffers-compiler/src/Orphan/Text.hs
```diff
@@ -0,0 +1,17 @@
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Orphan.Text () where
```
bladyjoker, line 3: Why not https://hackage.haskell.org/package/quickcheck-instances? Also base types are almost all sized afaik, did you run into some problems?

lambda-buffers-compiler/test/Test/Samples/Proto/Utils.hs
```diff
@@ -1,4 +1,4 @@
-module Test.Samples.Proto.Helpers (
+module Test.Samples.Proto.Utils (
```
bladyjoker, line 1: I'd like to have everything under test/Test/Utils folder. No need to say 'Samples/Proto' imo

lambda-buffers-compiler/test/Test/KindCheck.hs
```diff
@@ -46,14 +60,35 @@ kcTestFailing =
     assertBool "Test should have failed." $
       check_ compilerInput'incoherent /= Right ()
 
+{- | TyDef order does not matter when kind checking.
+
+ We're not interested in the failure error as there might be more than two
+ errors in a module - and it is non-determistic which one is first. But it is
```
bladyjoker, line 66: You could just use Set on errors and the `Eq` on them ignores the order.

bladyjoker (commented):
Just some nits for ya!

A general topic tho is where to place Arbitrary instances, I prefer all arbitraries orphaned in a separate module thus keeping the type defs clean and the Arbitrary instance colocated (convenient because the gen patterns are usually similar).

lambda-buffers-compiler/test/Test/KindCheck.hs
cstml: Not really - the compiler throws one error and it stops. That would mean that the compiler would have to find all errors - which is not what it is doing atm.

lambda-buffers-compiler/src/Orphan/Text.hs
cstml: Yes - my tests were OOM'ing - see the conversation on slack. The sized instances were what made it stop.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/Types.hs
cstml: I guess it's a matter of taste. I like having instances next to their definition - I find it's the nicest way especially when navigating library code. There's a second option where they go after the type definitions.

I'm not a fan of orphaned definitions - to me it _feels_ wrong. I'm also unsure how that affects the haddock output. 

I'd keep them in the file - if you want we can move them at the bottom. I would not make them orphan.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/Types.hs
cstml: This was one of the terms that was OOM'ing. It needs to be sized otherwise it recurses to the depth of _Mordor_.

bladyjoker (approved):

PR #35, opened 2023-02-02
Frontend: Integration with the Compiler

bladyjoker:
DONE
- `LambdaBuffers.Frontend.ToProto` module that maps the Frontend syntax into the `CompilerInput`
- Upgraded the Frontend CLI to invoke the Compiler CLI
- Improved Compiler testing a bit
- Renamed clis into `lbc` and `lbf`


lambda-buffers-compiler/test/Test/LambdaBuffers/Compiler/Gen.hs
```diff
@@ -73,27 +74,44 @@ genSum args = do
...
+
+genTyVar :: NonEmpty TyArg -> QC.Gen Ty
+genTyVar args = do
+  ar <- QC.elements (toList args)
   return $ defMessage & tyVar . varName .~ (ar ^. argName)
 
 genConstructor :: [TyArg] -> ConstrName -> QC.Gen Sum'Constructor
 genConstructor args cn = do
-  tys <- QC.chooseInt (1, limit) >>= vecOf (genTy args)
+  tys <- QC.chooseInt (1, limit) >>= vecOf (genTy [] args)
```
cstml, line 95:
```suggestion
  tys <- QC.chooseInt (1, limit) >>= vecOf (genTy mempty args)
```

lambda-buffers-frontend/app/LambdaBuffers/Frontend/Cli/Compile.hs
```diff
@@ -1,15 +1,27 @@
...
+import System.Process (callProcess)
 
 data CompileOpts = CompileOpts
   { _importPaths :: [FilePath]
   , _moduleFilepath :: FilePath
+  , _compilerCliFilepath :: FilePath
+  , _debug :: Bool
+  , _workingDir :: Maybe FilePath
   }
   deriving stock (Eq, Show)
```
cstml, line 27: It would be good to define a Monad for this at this point - otherwise the more we keep it as functions on IO the harder it will become later on to add other Effects like `Katip` etc. With that said - worth refactoring to a `Reader CompileOpts`

bladyjoker: Hmm, I don't think it's worth it as this shouldn't grow much more. I'm comfie with keeping this bare for now and we can ofc refactor it into a monad later on if necessary.

lambda-buffers-frontend/src/LambdaBuffers/Frontend.hs
```diff
@@ -70,12 +75,13 @@ instance Show FrontendError where
...
-runFrontend :: MonadIO m => [FilePath] -> FilePath -> m (Either FrontendError (Map (ModuleName ()) (Module SourceInfo)))
+runFrontend :: MonadIO m => [FilePath] -> FilePath -> m (Either FrontendError FrontendResult)
 runFrontend importPaths modFp = do
-  let stM = runReaderT (processFile modFp) (FrontRead (ModuleName [] undefined) [] importPaths)
+  let stM = runReaderT (processFile modFp) (FrontRead (ModuleName [] (SourceInfo "" (SourcePos 0 0) (SourcePos 0 0))) [] importPaths)
       exM = runStateT stM (FrontState mempty)
       ioM = runExceptT exM
-  fmap (importedModules . snd) <$> ioM
+  _ <- ioM
+  fmap (FrontendResult . importedModules . snd) <$> ioM
```
cstml, line 84:
```suggestion
  fmap (FrontendResult . importedModules . snd) <$> runExceptT exM
```

cstml (approved):

lambda-buffers-frontend/resources/good/work-dir/compiler-input.textproto
```diff
@@ -0,0 +1,979 @@
+modules {
```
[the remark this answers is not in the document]
bladyjoker: I'm keeping them around because I need something to point at. In the following work, I'll structure these as goldens and use them in tests (currently these 2 are not used in tests)

Issue #42, opened 2023-02-08
TODO(@gnumonik): This will need to use API types, to perform the detection and report back (TyClassDef, [ClassRef]).


gnumonik:
TODO(@gnumonik): This will need to use API types, to perform the detection and report back (TyClassDef, [ClassRef]).

_Originally posted by @bladyjoker in https://github.com/mlabs-haskell/lambda-buffers/pull/36#discussion_r1099830967_

Issue #41, opened 2023-02-08
 Expose the ClassCycleError in the API and report it.


gnumonik:
TODO(@gnumonik): Expose the ClassCycleError in the API and report it.

_Originally posted by @bladyjoker in https://github.com/mlabs-haskell/lambda-buffers/pull/36#discussion_r1099828919_

Issue #40, opened 2023-02-08
`supers` should be `ClassConstraint` as they seem fundamentally a different concept.


gnumonik:
TODO(@gnumonik): `supers` should be `ClassConstraint` as they seem fundamentally a different concept.

_Originally posted by @bladyjoker in https://github.com/mlabs-haskell/lambda-buffers/pull/36#discussion_r1099832184_

bladyjoker:
Done

PR #43, opened 2023-02-09
compiler/typeclass-solver

gnumonik:
This PR contains the solver (rewritten to avoid use of flip), a few small changes in response to comments from the last PR, and a suite of tests for the solver. 

@bladyjoker 
 - #41 / #42 can be done in the next PR. I'll have to rewrite the cycle detection *again* once I have the stuff from `Utils` in, there's not much of a point in doing it here only to redo it again tomorrow 
 - #40 is a proto-level change that would require a bunch of rewrites and seems out of place here (you told me to keep the PRs small). I'll do it in a standalone PR either before or after I get `Utils` & co merged

### DIFF
--- a/lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Pat.hs
+++ b/lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Pat.hs
@@ -19,10 +19,14 @@ cost of significantly more complex type signatures.
 -}
 
 data Pat
-  = {- extremely stupid, unfortunately necessary -}
+  = {- Name / ModuleName / Opaque / TyVarP are literal patterns (or ground terms)
+       because hey cannot contain any VarPs and therefore "have no holes".
+       Every TyDef or subcomponent thereof will be translated into a composite
+       pattern "without any holes". (Nil is also a literal/ground term, I guess) -}
     Name Text
-  | ModuleName [Text] -- also stupid, also necessary -_-
+  | ModuleName [Text]
   | Opaque
+  | TyVarP Text
   | {- Lists (constructed from Nil and :*) with bare types are used to
        encode products (where a list of length n encodes an n-tuple)
        Lists with field labels (l := t) are used to encode records and sum types
@@ -45,10 +49,8 @@ data Pat
   | ProdP Pat {- Pat arg should be a list of "Bare types" -}
   | SumP Pat {- where the Pat arg is expected to be (Constr l t :* rest) or Nil, where
                 rest is either Nil or a tyList of Constrs -}
-  | VarP Text {- This isn't a type variable. Although it is used to represent them in certain contexts,
-                 it is also used more generally to refer to any "hole" in a pattern to which another pattern
-                 may be substituted. We could have separate constr for type variables but it doesn't appear to be
-                 necessary at this time. -}
+  | VarP Text {- This isn't a type variable. It is used more generally to refer to any "hole" in a pattern into
+                 to which another pattern may be substituted. TyVarP is the literal pattern / ground term for TyVars  -}
   | RefP Pat Pat {- 1st arg should be a ModuleName  -}
   | AppP Pat Pat {- Pattern for Type applications -}
   | {- This last one is a bit special. This represents a complete type declaration.
--- a/lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Rules.hs
+++ b/lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Rules.hs
@@ -5,16 +5,22 @@ module LambdaBuffers.Compiler.TypeClass.Rules (
   Class (..),
   Constraint (..),
   Rule (..),
-  type Instance,
+  FQClassName (..),
   mapPat,
+  ruleHeadPat,
+  ruleHeadClass,
 ) where
 
-import LambdaBuffers.Compiler.ProtoCompat qualified as P
+import Data.Text (Text)
+
 import LambdaBuffers.Compiler.TypeClass.Pat (Pat)
 
+data FQClassName = FQClassName {cName :: Text, cModule :: [Text]}
+  deriving stock (Show, Eq, Ord)
+
 data Class = Class
-  { name :: P.TyClassRef
-  , supers :: [Class]
+  { cname :: FQClassName
+  , csupers :: [Class]
   }
   deriving stock (Show, Eq, Ord)
 
@@ -33,9 +39,15 @@ data Rule where
   deriving stock (Show, Eq, Ord)
 infixl 7 :<=
 
-type Instance = Rule
-
 {- Map over the Pats inside of an Rule
 -}
 mapPat :: (Pat -> Pat) -> Rule -> Rule
 mapPat f (C c ty :<= is) = C c (f ty) :<= map (\(C cx p) -> C cx (f p)) is
+
+{- Extract the inner Pat from a Rule head
+-}
+ruleHeadPat :: Rule -> Pat
+ruleHeadPat (C _ p :<= _) = p
+
+ruleHeadClass :: Rule -> Class
+ruleHeadClass (C c _ :<= _) = c
--- a/lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
+++ b/lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
@@ -1,0 +1,145 @@
+{-# LANGUAGE LambdaCase #-}
+
+module LambdaBuffers.Compiler.TypeClass.Solve (solve) where
+
+import Data.List (foldl', sortBy)
+import Data.Text (Text)
+import LambdaBuffers.Compiler.TypeClass.Pat (
+  Pat (AppP, DecP, ProdP, RecP, RefP, SumP, VarP, (:*), (:=)),
+  matches,
+ )
+import LambdaBuffers.Compiler.TypeClass.Rules (
+  Class (csupers),
+  Constraint (C),
+  Rule ((:<=)),
+  mapPat,
+  ruleHeadClass,
+  ruleHeadPat,
+ )
+
+import Data.Set qualified as S
+
+{- Variable substitution. Given a string that represents a variable name,
+   and a type to instantiate variables with that name to, performs the
+   instantiation
+-}
+subV :: Text -> Pat -> Pat -> Pat
+subV varNm t = \case
+  var@(VarP v) -> if v == varNm then t else var
+  x :* xs -> subV varNm t x :* subV varNm t xs
+  l := x -> subV varNm t l := subV varNm t x
+  ProdP xs -> ProdP (subV varNm t xs)
+  RecP xs -> RecP (subV varNm t xs)
+  SumP xs -> SumP (subV varNm t xs)
+  AppP t1 t2 -> AppP (subV varNm t t1) (subV varNm t t2)
+  RefP n x -> RefP (subV varNm t n) (subV varNm t x)
+  DecP a b c -> DecP (subV varNm t a) (subV varNm t b) (subV varNm t c)
+  other -> other
+
+{- Performs substitution on an entire instance (the first argument) given the
+   concrete types from a Pat (the second argument).
+   Note that ONLY PatVars which occur in the Instance *HEAD* are replaced, though they
+   are replaced in the instance superclasses as well (if they occur there).
+-}
+subst :: Rule -> Pat -> Rule
+subst cst@(C _ t :<= _) ty = mapPat (go (getSubs t ty)) cst
+  where
+    go :: [(Text, Pat)] -> Pat -> Pat
+    go subs tty =
+      let noflip p1 p2 = uncurry subV p2 p1
+       in foldl' noflip tty subs
+
+{- Given two patterns (which are hopefully structurally similar), gather a list of all substitutions
+   from the PatVars in the first argument to the concrete types (hopefully!) in the second argument
+-}
+getSubs :: Pat -> Pat -> [(Text, Pat)] -- should be a set, whatever
+getSubs (VarP s) t = [(s, t)]
+getSubs (x :* xs) (x' :* xs') = getSubs x x' <> getSubs xs xs'
+getSubs (l := t) (l' := t') = getSubs l l' <> getSubs t t'
+getSubs (ProdP xs) (ProdP xs') = getSubs xs xs'
+getSubs (RecP xs) (RecP xs') = getSubs xs xs'
+getSubs (SumP xs) (SumP xs') = getSubs xs xs'
+getSubs (AppP t1 t2) (AppP t1' t2') = getSubs t1 t1' <> getSubs t2 t2'
+getSubs (RefP n t) (RefP n' t') = getSubs n n' <> getSubs t t'
+getSubs (DecP a b c) (DecP a' b' c') = getSubs a a' <> getSubs b b' <> getSubs c c'
+getSubs _ _ = []
+
+-- should be vastly more efficient than Data.List.Nub
+deduplicate :: Ord a => [a] -> [a]
+deduplicate = S.toList . S.fromList
+
+-- is the first pattern a substitution instance of the second
+isSubstitutionOf :: Pat -> Pat -> Bool
+isSubstitutionOf p1 p2 = matches p2 p1
+
+compareSpecificity :: Pat -> Pat -> Ordering
+compareSpecificity p1 p2
+  | p1
+      `isSubstitutionOf` p2
+      && p2
+      `isSubstitutionOf` p1 =
+      EQ
+  | p1 `isSubstitutionOf` p2 = LT
+  | otherwise = GT
+
+sortOnSpecificity :: Pat -> Class -> [Rule] -> [Rule]
+sortOnSpecificity p c ps =
+  sortBy (\a1 a2 -> compareSpecificity (ruleHeadPat a1) (ruleHeadPat a2)) $
+    filter matchPatAndClass ps
+  where
+    matchPatAndClass :: Rule -> Bool
+    matchPatAndClass r =
+      ruleHeadClass r == c
+        && ruleHeadPat r
+        `matches` p
+
+mostSpecificInstance :: Pat -> Class -> [Rule] -> Maybe Rule
+mostSpecificInstance p c ps = case sortOnSpecificity p c ps of
+  [] -> Nothing
+  (x : _) -> Just x
+
+{- Given a list of instances (the initial scope), determines whether we can derive
+   an instance of the Class argument for the Pat argument. A result of [] indicates that there are
+   no remaining subgoals and that the constraint has been solved.
+   NOTE: At the moment this handles superclasses differently than you might expect -
+         instead of assuming that the superclasses for all in-scope classes are defined,
+         we check that those constraints can be solved before affirmatively judging that the
+         target constraint has been solved. I *think* that makes sense in this context (whereas in Haskell
+         it doesn't b/c it's *impossible* to have `instance Foo X` if the definition of Foo is
+         `class Bar y => Foo y` without an `instance Bar X`)
+-}
+solve ::
+  [Rule] -> -- all instance rules in scope. WE ASSUME THESE HAVE ALREADY BEEN GENERATED, SOMEWHERE
+  Constraint -> -- constraint we're trying to solve
+  [Constraint] -- subgoals that cannot be solved for w/ the current rule set
+solve inScope cst@(C c pat) =
+  -- First, we look for the most specific instance...
+  case mostSpecificInstance pat c inScope of
+    -- If there isn't one, we return only the constraint we were trying to solve
+    Nothing -> [cst]
+    -- If there is, we substitute the argument of the constraint to be solved into the matching rules
+    Just rule -> case subst rule pat of
+      -- If there are no additional constraints on the rule, we try to solve the superclasses
+      C _ p :<= [] -> case csupers c of
+        [] -> []
+        xs -> solveClassesFor p xs
+      -- If there are additional constraints on the rule, we try to solve them
+      C _ _ :<= is -> case concatMap (solve inScope) is of
+        -- If we succeed at solving the additional constraints on the rule, we try to solve the supers
+        [] -> solveClassesFor pat (csupers c)
+        -- We deduplicate the list of unsolvable subgoals
+        xs -> deduplicate xs
+  where
+    -- NOTE(@bladyjoker): The version w/ flip is more performant...
+    -- Given a Pat and a list of Classes, attempt to solve the constraints
+    -- constructed from the Pat and each Class
+    solveClassesFor :: Pat -> [Class] -> [Constraint]
+    solveClassesFor p =
+      deduplicate -- multiple constraints could emit the same subgoal which is bad
+        . concatMap
+          ( solve inScope
+              . ( \cls ->
+                    let classConstraint = C cls p
+                     in classConstraint
+                )
+          )
--- a/lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
+++ b/lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
@@ -1,9 +1,7 @@
 {-# LANGUAGE LambdaCase #-}
 
-module LambdaBuffers.Compiler.TypeClass.Solve (solve) where
+module LambdaBuffers.Compiler.TypeClass.Solve (solveM, solve, Overlap (..)) where
 
-import Data.List (foldl', sortBy)
-import Data.Text (Text)
 import LambdaBuffers.Compiler.TypeClass.Pat (
   Pat (AppP, DecP, ProdP, RecP, RefP, SumP, VarP, (:*), (:=)),
   matches,
@@ -17,11 +15,20 @@ import LambdaBuffers.Compiler.TypeClass.Rules (
   ruleHeadPat,
  )
 
+import Control.Monad.Except (throwError)
+import Control.Monad.Reader (ReaderT, runReaderT)
+import Control.Monad.Reader.Class (MonadReader (ask))
+import Control.Monad.Writer.Class (MonadWriter (tell))
+import Control.Monad.Writer.Strict (WriterT, execWriterT)
+import Data.Foldable (traverse_)
+import Data.List (foldl')
 import Data.Set qualified as S
+import Data.Text (Text)
 
-{- Variable substitution. Given a string that represents a variable name,
-   and a type to instantiate variables with that name to, performs the
-   instantiation
+{- Pattern/Template/Unification variable  substitution.
+   Given a string that represents a variable name,
+   and a type to instantiate variables with that name to,
+   performs the instantiation
 -}
 subV :: Text -> Pat -> Pat -> Pat
 subV varNm t = \case
@@ -64,28 +71,20 @@ getSubs (RefP n t) (RefP n' t') = getSubs n n' <> getSubs t t'
 getSubs (DecP a b c) (DecP a' b' c') = getSubs a a' <> getSubs b b' <> getSubs c c'
 getSubs _ _ = []
 
--- should be vastly more efficient than Data.List.Nub
-deduplicate :: Ord a => [a] -> [a]
-deduplicate = S.toList . S.fromList
+-- NoMatch isn't fatal but OverlappingMatches is (i.e. we need to stop when we encounter it)
+data MatchError
+  = NoMatch
+  | OverlappingMatches [Rule]
 
--- is the first pattern a substitution instance of the second
-isSubstitutionOf :: Pat -> Pat -> Bool
-isSubstitutionOf p1 p2 = matches p2 p1
+-- for SolveM, since we catch NoMatch
+data Overlap = Overlap Constraint [Rule]
+  deriving stock (Show, Eq)
 
-compareSpecificity :: Pat -> Pat -> Ordering
-compareSpecificity p1 p2
-  | p1
-      `isSubstitutionOf` p2
-      && p2
-      `isSubstitutionOf` p1 =
-      EQ
-  | p1 `isSubstitutionOf` p2 = LT
-  | otherwise = GT
-
-sortOnSpecificity :: Pat -> Class -> [Rule] -> [Rule]
-sortOnSpecificity p c ps =
-  sortBy (\a1 a2 -> compareSpecificity (ruleHeadPat a1) (ruleHeadPat a2)) $
-    filter matchPatAndClass ps
+selectMatchingInstance :: Pat -> Class -> [Rule] -> Either MatchError Rule
+selectMatchingInstance p c rs = case filter matchPatAndClass rs of
+  [] -> Left NoMatch
+  [r] -> Right r
+  overlaps -> Left $ OverlappingMatches overlaps
   where
     matchPatAndClass :: Rule -> Bool
     matchPatAndClass r =
@@ -93,10 +92,7 @@ sortOnSpecificity p c ps =
         && ruleHeadPat r
         `matches` p
 
-mostSpecificInstance :: Pat -> Class -> [Rule] -> Maybe Rule
-mostSpecificInstance p c ps = case sortOnSpecificity p c ps of
-  [] -> Nothing
-  (x : _) -> Just x
+type SolveM = ReaderT [Rule] (WriterT (S.Set Constraint) (Either Overlap))
 
 {- Given a list of instances (the initial scope), determines whether we can derive
    an instance of the Class argument for the Pat argument. A result of [] indicates that there are
@@ -108,38 +104,28 @@ mostSpecificInstance p c ps = case sortOnSpecificity p c ps of
          it doesn't b/c it's *impossible* to have `instance Foo X` if the definition of Foo is
          `class Bar y => Foo y` without an `instance Bar X`)
 -}
-solve ::
-  [Rule] -> -- all instance rules in scope. WE ASSUME THESE HAVE ALREADY BEEN GENERATED, SOMEWHERE
-  Constraint -> -- constraint we're trying to solve
-  [Constraint] -- subgoals that cannot be solved for w/ the current rule set
-solve inScope cst@(C c pat) =
-  -- First, we look for the most specific instance...
-  case mostSpecificInstance pat c inScope of
-    -- If there isn't one, we return only the constraint we were trying to solve
-    Nothing -> [cst]
-    -- If there is, we substitute the argument of the constraint to be solved into the matching rules
-    Just rule -> case subst rule pat of
-      -- If there are no additional constraints on the rule, we try to solve the superclasses
-      C _ p :<= [] -> case csupers c of
-        [] -> []
-        xs -> solveClassesFor p xs
-      -- If there are additional constraints on the rule, we try to solve them
-      C _ _ :<= is -> case concatMap (solve inScope) is of
-        -- If we succeed at solving the additional constraints on the rule, we try to solve the supers
-        [] -> solveClassesFor pat (csupers c)
-        -- We deduplicate the list of unsolvable subgoals
-        xs -> deduplicate xs
+solveM :: Constraint -> SolveM ()
+solveM cst@(C c pat) =
+  ask >>= \inScope ->
+    -- First, we look for the most specific instance...
+    case selectMatchingInstance pat c inScope of
+      Left e -> case e of
+        NoMatch -> tell $ S.singleton cst
+        OverlappingMatches olps -> throwError $ Overlap cst olps
+      -- If there is, we substitute the argument of the constraint to be solved into the matching rules
+      Right rule -> case subst rule pat of
+        -- If there are no additional constraints on the rule, we try to solve the superclasses
+        C _ p :<= [] -> solveClassesFor p (csupers c)
+        -- If there are additional constraints on the rule, we try to solve them
+        C _ _ :<= is -> do
+          traverse_ solveM is
+          solveClassesFor pat (csupers c)
   where
     -- NOTE(@bladyjoker): The version w/ flip is more performant...
     -- Given a Pat and a list of Classes, attempt to solve the constraints
     -- constructed from the Pat and each Class
-    solveClassesFor :: Pat -> [Class] -> [Constraint]
-    solveClassesFor p =
-      deduplicate -- multiple constraints could emit the same subgoal which is bad
-        . concatMap
-          ( solve inScope
-              . ( \cls ->
-                    let classConstraint = C cls p
-                     in classConstraint
-                )
-          )
+    solveClassesFor :: Pat -> [Class] -> SolveM ()
+    solveClassesFor p = traverse_ (\cls -> solveM (C cls p))
+
+solve :: [Rule] -> Constraint -> Either Overlap [Constraint]
+solve rules c = fmap S.toList $ execWriterT $ runReaderT (solveM c) rules
--- a/lambda-buffers-compiler/test/Test/TypeClassCheck.hs
+++ b/lambda-buffers-compiler/test/Test/TypeClassCheck.hs
@@ -1,6 +1,6 @@
 {-# LANGUAGE PatternSynonyms #-}
 
-module Test.TypeClassCheck (test) where
+module Test.TypeClassCheck {- (test) -} where
 
 import Control.Lens ((.~))
 import Data.Function ((&))
@@ -19,6 +19,7 @@ import LambdaBuffers.Compiler.TypeClass.Pat (
     RecP,
     RefP,
     SumP,
+    TyVarP,
     VarP,
     (:*),
     (:=)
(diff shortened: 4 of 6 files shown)


lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Rules.hs
```diff
@@ -5,16 +5,22 @@ module LambdaBuffers.Compiler.TypeClass.Rules (
...
+  ruleHeadClass,
 ) where
 
-import LambdaBuffers.Compiler.ProtoCompat qualified as P
+import Data.Text (Text)
+
 import LambdaBuffers.Compiler.TypeClass.Pat (Pat)
 
+data FQClassName = FQClassName {cName :: Text, cModule :: [Text]}
+  deriving stock (Show, Eq, Ord)
```
bladyjoker, line 19: I think the reason you need this is because your `solve` doesn't have access to any 'context'. That's ok, but introduces separate processing steps which is not really necessary.

TyClassRef would be sufficient if your solver worked with
 ```haskell
 SolveContext :: '[ "currentModule" := ModuleName, "classDefs" :=  Map TyClassName TyClassDef]
 ```

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Rules.hs
```diff
@@ -5,16 +5,22 @@ module LambdaBuffers.Compiler.TypeClass.Rules (
...
+
 data Class = Class
-  { name :: P.TyClassRef
-  , supers :: [Class]
+  { cname :: FQClassName
+  , csupers :: [Class]
   }
```
bladyjoker, line 24: Same comment as above. No need to change it, but consider the alternative approach with context building.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
```diff
@@ -0,0 +1,145 @@
...
+  var@(VarP v) -> if v == varNm then t else var
+  x :* xs -> subV varNm t x :* subV varNm t xs
+  l := x -> subV varNm t l := subV varNm t x
+  ProdP xs -> ProdP (subV varNm t xs)
+  RecP xs -> RecP (subV varNm t xs)
+  SumP xs -> SumP (subV varNm t xs)
+  AppP t1 t2 -> AppP (subV varNm t t1) (subV varNm t t2)
+  RefP n x -> RefP (subV varNm t n) (subV varNm t x)
+  DecP a b c -> DecP (subV varNm t a) (subV varNm t b) (subV varNm t c)
+  other -> other
```
bladyjoker, line 37: Imo this seems very odd. Variable substitution should ONLY happen when you perform a type application. I'm not convinced this is not a hack of some kind...

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
```diff
@@ -0,0 +1,145 @@
...
+   Note that ONLY PatVars which occur in the Instance *HEAD* are replaced, though they
+   are replaced in the instance superclasses as well (if they occur there).
+-}
+subst :: Rule -> Pat -> Rule
+subst cst@(C _ t :<= _) ty = mapPat (go (getSubs t ty)) cst
+  where
+    go :: [(Text, Pat)] -> Pat -> Pat
+    go subs tty =
+      let noflip p1 p2 = uncurry subV p2 p1
+       in foldl' noflip tty subs
```
bladyjoker, line 50: This is very much unification.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
```diff
@@ -0,0 +1,145 @@
...
+   no remaining subgoals and that the constraint has been solved.
+   NOTE: At the moment this handles superclasses differently than you might expect -
+         instead of assuming that the superclasses for all in-scope classes are defined,
+         we check that those constraints can be solved before affirmatively judging that the
+         target constraint has been solved. I *think* that makes sense in this context (whereas in Haskell
+         it doesn't b/c it's *impossible* to have `instance Foo X` if the definition of Foo is
+         `class Bar y => Foo y` without an `instance Bar X`)
+-}
+solve ::
+  [Rule] -> -- all instance rules in scope. WE ASSUME THESE HAVE ALREADY BEEN GENERATED, SOMEWHERE
```
bladyjoker, line 112: I would add turn this into a SolverContext and add all the rules there.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
```diff
@@ -0,0 +1,145 @@
...
+  Constraint -> -- constraint we're trying to solve
+  [Constraint] -- subgoals that cannot be solved for w/ the current rule set
+solve inScope cst@(C c pat) =
+  -- First, we look for the most specific instance...
+  case mostSpecificInstance pat c inScope of
```
bladyjoker, line 117: I need some clarity on this, you should be needing this because we don't have duplicate instances right?

lambda-buffers-compiler/test/Test/TypeClassCheck.hs
```diff
@@ -70,3 +101,219 @@ nocycles =
   , mkclass "Monad" ["Applicative"]
   , mkclass "Traversable" ["Foldable", "Functor"]
   ]
+
+---- Solver tests
+
+solveTests :: TestTree
+solveTests =
```
bladyjoker, line 108: Thanks for the tests.

lambda-buffers-compiler/test/Test/TypeClassCheck.hs
```diff
@@ -70,3 +101,219 @@ nocycles =
...
+pattern NoConstraints :: Class -> Pat -> Rule
+pattern NoConstraints c p = C c p :<= []
+
+pattern Labeled :: Text -> Pat -> Pat
+pattern Labeled nm p = Name nm := p
+
+userRules1 :: Class -> [Rule]
+userRules1 c =
+  [ NoConstraints c Int
+  , NoConstraints c Bool
```
bladyjoker, line 198: It's nicer with just `C c Int :<= []` imo

lambda-buffers-compiler/test/Test/TypeClassCheck.hs
```diff
@@ -70,3 +101,219 @@ nocycles =
...
+  , C c (Maybe _x) :<= [C c _x]
+  , C c (Either _l _x) :<= [C c _l, C c _x]
+  , C c (List _x) :<= [C c _x]
```
bladyjoker, line 201: I don't think this does what we need it to do...

Consider...

```haskell

-- Int and String don't have an Eq

data Foo a = MkFoo Int String (Maybe a)

instance Eq a => Eq (Maybe a)

instance Eq a => Eq (Foo a)
```

lambda-buffers-compiler/test/Test/TypeClassCheck.hs
```diff
@@ -70,3 +101,219 @@ nocycles =
...
+mkStructuralRules :: Class -> [Rule]
+mkStructuralRules c =
+  [ C c Nil :<= []
+  , C c (_x :* _xs) :<= [C c _x, C c _xs]
+  , C c (_l := _x) :<= [C c _x]
+  , C c (RecP _xs) :<= [C c _xs]
+  , C c (ProdP _xs) :<= [C c _xs]
+  , C c (SumP _xs) :<= [C c _xs]
+  , C c (DecP _name _vars _body) :<= [C c _body]
+  ]
```
bladyjoker, line 166: Imo our structural rules for types of kind `*` should be

```prolog
% canonical representation of TyBody
class(unit).
class(void).
class(either(A,B)) :- class(A), class(B).
class(prod(A,B)) :- class(A), class(B).

% type application
class(ty_app(F, A)) :- apply(F, A, Res), class(Res).

% ty_ref
class(ty_ref(Ref)) :- ty_def(Ref, Ty), class(Ty).

% and perhaps for 0 arg abstractions
class(ty_abs([], TyBody)) :- class(TyBody).
```
There are 2 special rules that deal with lambda application (unwrapping the type def) and fetching a ty ref from the context.

bladyjoker: Read on how that affects things.

lambda-buffers-compiler/test/Test/TypeClassCheck.hs
bladyjoker: What happens in your implementation?

bladyjoker: Which rules fire?

bladyjoker (commented):
Let's do some more talking about this, but thank you this is a very clean PR! Good job

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
bladyjoker: Also can we please run the solver with a Reader context and a Writer one to collect the solver trace? That would be extremely useful!

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Rules.hs
gnumonik: `solve` isn't supposed to have access to a context. It's supposed to be the simplest possible implementation of the constraint solver algorithm, which is used as the central component later on when a context (or something that's effectively the same) gets constructed. 

There's no hard theoretical reason for that - `solve` was just hard to debug so I cut it down to the bare essentials. 

In my mind, `FQClassName` was just a safety valve to make sure we don't waste time creating/working with classes that differ only in the `SourceInfo` of their `TyClassRef`. I guess I could use `TyClassRef` and a slightly degenerate Eq instance, /shrug

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
gnumonik: I really don't think it is. Unification is more than just substitution. I'm not doing unification if I have a logical sentence 

```
P -> (P v Q)
```
and I substitute `R & S` into `P` yielding 

```
(R&S) -> ((R&S) v Q)
```

gnumonik: You have to understand what the point of this is. In practice the ONLY point of `subst` is to substitute into type variables in the "constraints" part of a `Rule`. We do it to the head as well because... well, we might as well I guess, but it's always a unidirectional thing. 

If the head of a rule matches a `Pat` then we don't actually care about the head, but we need to make sure the pattern variables in the constraints are instantiated to the right thing

gnumonik: Really this *can't* be unification.... if for no other reason than this was well established 100+ of years before unification existed. If this were unification then Herbrand didn't discover anything when he discovered unification lol

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
gnumonik: Er, yeah, we can do that.

lambda-buffers-compiler/test/Test/TypeClassCheck.hs
gnumonik: yeah you're right i got carried away w/ pattern synonyms

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
gnumonik: I won't spend forever on this since it doesn't really matter much, but there's a way of understanding how this works that looks a lot more like case expressions than like unification. The `Rule`s can be thought of as patterns (in the Haskell sense, the thing that goes to the left of `->` in the body of a case expression) and the `Pat` in the head of the constraint we're solving for can be thought of as an expression. 

So if we have 

```
scope: 
rule C Int 
rule C a => C (Maybe a)
```

what that "means" is something like 

```
solve x = case x of 
  C Int -> ()                                             -- Branch A 
  C (Maybe patVar) -> solve patVar   -- Branch B 
```

So, concretely w/ `C (Maybe Int)`: 

```
solve (C  (Maybe Int))
   > solve Int                     -- matches Branch B 
   > ()                                  -- matches Branch A 
```

And substitution (which is always substituting a `C (Maybe Int)` _expression-ey_ thing into a more general pattern w/ a patVar "hole" like `C (Maybe a)`) is just how we determine whether something matches the pattern. `C Int` is effectively a literal pattern that matches one specific expression.

gnumonik: Or, to try another way of putting it: It's not unification b/c the solution to a unification problem involves a uniform assignment of terms to variables _on both sides_ of the equation. But we only "assign" terms to variables on the left hand (more general) side. 

If I have a rule `C (F a)` and I'm trying to solve `C (F (G b))`, I'll "assign" `F (G b)` to `a` but I'll **NEVER** assign anything to `b`. 

We are assigning variables uniformly to both sides of the `=>` in a rule, but `=>` doesn't express an equality so even there we're just doing substitution.

gnumonik: c.f. [this better explanation of the difference between pattern matching and unification](https://pages.cpsc.ucalgary.ca/~robin/class/449/notes/MatchUnif.pdf)

Variables on the "left hand side" (i.e. in Rules) are what you might call "unification variables" but there are no unification variables on the right hand side. If one side by definition (or artifice in this case) doesn't have unification variables then we're doing pattern matching.

...which is admittedly somewhat obscured by my implementation here since `VarP` on the Rule side means something different than `VarP` on the right hand side.

gnumonik: And, therefore, `subst` doesn't do unification. `Rule`s are always the "template" and bare `Pat`s are always the "subject". There _aren't_ "templates" and "subjects" in Prolog the way there are in pattern matching expressions in Haskell (or other languages).

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
gnumonik: Honestly I was doing it because that's what the research I did lead me to believe I should do. 

The problem is really overlapping instances, not duplicates (the machinery that prepares this uses `Set Rule` so there's not even a point in worrying about duplicates here). 

After thinking about this, though, you're right that `mostSpecificInstance` doesn't really make much sense. If there are multiple matching instances then we have overlapping instances which we can't have. I'll revert that to something like the old implementation today, which threw an error if multiple in-scope instances match a given head. 

(There might be a place for overlapping instances if we wanted to express more sophisticated structural rules, and we may need them in a limited capacity to do things with `Nat`s if we ever implement those)

lambda-buffers-compiler/test/Test/TypeClassCheck.hs
gnumonik: Are we trying to derive both of those instances or are we assuming them as rules? 

If we try to derive them, we have to "pretend" that we have `Eq a` in scope (i.e. we assume it). But - and I'm only just realizing this now - what we specifically need to assume is that `Eq a` holds for the _type variable_ `a` and not the _unification variable_ `a`. So, as I hinted at above, I need to distinguish `VarP` from something like `TyVarP` (where `TyVarP "a"` is a _literal pattern_ or ground term if you like). 

This is the correct thing to assume for codegen because when writing the methods for an  `instance Eq a => Eq (Foo a)` , the compiler assumes that `a` is an `Eq` instance. (Once you try to write these you'll realize that all the codegen for the instances looks the same, we're just trying to figure out whether it's OK to generate the code or whether we should report an error and stop prior to code generation) 

`solve` isn't meant to be the sum total of the derivation machinery, and what exactly is in scope depends upon which of the instances we're trying to derive. Let's pretend we're trying to derive `Eq a => Eq (Foo a)` b/c that's more interesting: 
 
Let's convert to a pattern 

```
foo = DecP (Name "Foo) (Name "a" :* Nil)  $ SumP (Name "MkFoo" := ProdP ( Int :* String :* Maybe (TyVarP "a") :* Nil)) :* Nil   
```

Nothing fires w/o the structural rules (obviously) but assume we have those in scope, instances for `Int/String` in scope, and `Eq a` (where `a` is a literal `TyVarP "a"`, so really `Eq (TyVarP "a")` ) as a consequence of the above assumption procedure.  We also have `Eq a => Eq (Maybe a)` in scope here.  So our scope is something like: 

``` 
-- structural 
C Eq Nil :<= []
C Eq (_x :* _xs) :<= [C Eq _x, C Eq _xs]
C Eq (_l := _x) :<= [C Eq _x]
C Eq (RecP _xs) :<= [C Eq _xs]
C Eq (ProdP _xs) :<= [C Eq _xs]
C Eq (SumP _xs) :<= [C Eq _xs]
C Eq (DecP _name _vars _body) :<= [C Eq _body]
-- for the sake of the example 
C Eq Int 
C Eq String 
-- b/c we assume all other instances in the same module are in scope 
C Eq (Maybe _x) :<= [C Eq _x]
-- added assumption for constrained instances explained above 
C Eq (TyVarP "a")
```
Reduction:
```
-- via structural rule for DecP 
C (SumP (Name "mkFoo" := ProdP ( Int :* String :* Maybe (TyVarP "a") :* Nil)) :* Nil) 

-- via structural rules for SumP 
C ((Name "mkFoo" := ProdP ( Int :* String :* Maybe (TyVarP "a") :* Nil)) :* Nil)

-- via structural rule for Nil 
C (Name "mkFoo" := ProdP ( Int :* String :* Maybe (TyVarP "a") :* Nil)) 

-- via structural rule for (:=) 
C (ProdP ( Int :* String :* Maybe (TyVarP "a") :* Nil)

-- via structural rule for ProdP 
C ( Int :* String :* Maybe (TyVarP "a") :* Nil)

-- via structural rule for (:*) and (Eq Int)   - I'm starting at the head but we could do it the other way 
C (String :* Maybe (TyVarP "a") :* Nil)

-- via structural rule for (:*) and (Eq String) 
C (Maybe (TyVarP "a") :* Nil) 

-- via structural rules for :* and Nil 
C (Maybe (TyVarP "a") 

-- via rule for Maybe 
C (TyVarP "a") 

-- via assumption of (TyVarP "a") 
()
```

gnumonik: But do note that `solve` is just supposed to be the function you call _once you've already established the scope (including any additional assumptions)_. The process of establishing those assumptions is a separate concern.

gnumonik: @bladyjoker it's important that this is right, you can skim most of the other responses but pls read this one closely

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
gnumonik: See some of my remarks in other comments. I made a mistake. We need `TyVarP`. `VarP` should only represent unification variables. This isn't substitution on `TyVar`s, it's substitution on "unification vars" (which is a bad name, they're really  Pattern Variables, or Template Variables or something if you like)

This is 100% not a hack; I lifted the basic concept straight from the pattern matching machinery in Haskell (ask me for a link to the section of the Haskell report if you want it).

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Rules.hs
gnumonik: I'm indifferent here, what you're suggesting would work fine but would introduce the possibility of a new error type when we look up a class that should exist. There are already enough of those in `Utils` in the other PR.... but also we effectively construct the context you're proposing in order to construct all the classes so /shrug. 

I'm probably going to keep it as-is for now because it makes `solve` simpler but like I said, i have no strong feelings about this

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
gnumonik: I fixed this in the commits i'm about to push

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Rules.hs
bladyjoker: 1. No need to do the Eq
2. I told you SourceInfo should have been a type parameter :D
3. 'solve' does have access to a context, the first argument is a ruleset it can observe right? That's the context, I'm just saying it doesn't have to be the only thing in the context.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
```diff
@@ -0,0 +1,131 @@
...
+         instead of assuming that the superclasses for all in-scope classes are defined,
+         we check that those constraints can be solved before affirmatively judging that the
+         target constraint has been solved. I *think* that makes sense in this context (whereas in Haskell
+         it doesn't b/c it's *impossible* to have `instance Foo X` if the definition of Foo is
+         `class Bar y => Foo y` without an `instance Bar X`)
+-}
+solveM :: Constraint -> SolveM ()
+solveM cst@(C c pat) =
+  ask >>= \inScope ->
+    -- First, we look for the most specific instance...
```
bladyjoker, line 110: You can remove this comment.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
```diff
@@ -0,0 +1,131 @@
...
+    case selectMatchingInstance pat c inScope of
+      Left e -> case e of
+        NoMatch -> tell $ S.singleton cst
+        OverlappingMatches olps -> throwError $ Overlap cst olps
```
bladyjoker, line 114: See my comment below

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
```diff
@@ -0,0 +1,131 @@
...
+      -- If there is, we substitute the argument of the constraint to be solved into the matching rules
+      Right rule -> case subst rule pat of
+        -- If there are no additional constraints on the rule, we try to solve the superclasses
+        C _ p :<= [] -> solveClassesFor p (csupers c)
+        -- If there are additional constraints on the rule, we try to solve them
+        C _ _ :<= is -> do
+          traverse_ solveM is
+          solveClassesFor pat (csupers c)
```
bladyjoker, line 122: This reads nicer thanks.

I would strongly state that the `super class` constraints should be part of the rules themselves and not treat them specially. Specifically how I was doing it

```
ord(maybe(A)) :-
  ord(A),
  eq(maybe(A)).
```

And then the machinery just works as intended, no need to consider it specially right?

bladyjoker: That way you solver doesn't need any knowledge of what a superclass is, it's just a simple rule system.

lambda-buffers-compiler/test/Test/TypeClassCheck.hs
bladyjoker: Completely agree that your structural rules are identical to canonical ones, I just used them from brevity.

bladyjoker: The rules for TyRef and TyApp are essential, and them being absent is a signal we're a bit on the hacky side of things.

A typeclass Eq is declaring that it can be applied to types of kind `*`. What are those types?
1. Sum/Prod/Nil
2. TyApp - of kind *
3. TyRef - of kind *
4. TyAbs - of kind *

The things you're missing and should be added explicitly:

```prolog
% type application
class(ty_app(F, A)) :- apply(F, A, Res), class(Res).

% ty_ref
class(ty_ref(Ref)) :- ty_def(Ref, Ty), class(Ty).

% and perhaps for 0 arg abstractions
class(ty_abs([], TyBody)) :- class(TyBody).
```

What would `apply` and `ty_def` be in your formulation?

```haskell
data Rule where
  (:<=) :: Constraint -> [Constraint] -> Rule

data Constraint = C Class Pat | Apply Pat Pat Pat | TyDef TyRef Ty
```

You just extended your language to contain `apply` that performs 'lambda application' and `ty def` that just queries the 'context' for the requested Ty term bound to this TyRef. (which should fail btw with an API error)

I imagine that `solve` would then (i guess) use 'subV' the implement this call.

lambda-buffers-compiler/test/Test/TypeClassCheck.hs
bladyjoker: Let me make a better example!

1. Int has Eq
2. String doesn't have Eq

Let's consider both use cases, namely `derive` and `instance`:

```haskell
opaque Int
opaque String
data Foo a = MkFoo Int String (Maybe a)

instance Eq Int
instance Eq a => Eq (Maybe a)
instance Eq a => Eq (Foo a)
```
vs.

```haskell
opaque Int
opaque String
data Foo a = MkFoo Int String (Maybe a)

derive Eq Int
derive Eq (Maybe a)
derive Eq (Foo a)
```
In the `instance` case, this is a 'user supplied clause` and we just put it in amongst other rules:

```haskell
-- structural 
C Eq Nil :<= []
C Eq (_x :* _xs) :<= [C Eq _x, C Eq _xs]
C Eq (_l := _x) :<= [C Eq _x]
C Eq (RecP _xs) :<= [C Eq _xs]
C Eq (ProdP _xs) :<= [C Eq _xs]
C Eq (SumP _xs) :<= [C Eq _xs]
C Eq (DecP _name _vars _body) :<= [C Eq _body]

-- user specified
C Eq Int :<= []
C Eq (Maybe _x) :<= [C Eq _x]
C Eq (Foo _x) :<= [C Eq _x]
```

When I ask `Eq (Foo Int)` what happens? Given this rule set:

1. rule `C Eq (Foo _x) :<= [C Eq _x]` fires
2. rule `C Eq Int :<= []` fires
3. done --- which is correct as far as the Compiler is concerned right?

Unless I'm not seeing why necessarily `C Eq (DecP _name _vars _body) :<= [C Eq _body]` would fire??? The user can't really specify the rules we construct with `derive`.

However, in the `derive` case:

```haskell
-- structural 
C Eq Nil :<= []
C Eq (_x :* _xs) :<= [C Eq _x, C Eq _xs]
C Eq (_l := _x) :<= [C Eq _x]
C Eq (RecP _xs) :<= [C Eq _xs]
C Eq (ProdP _xs) :<= [C Eq _xs]
C Eq (SumP _xs) :<= [C Eq _xs]
C Eq (DecP _name _vars _body) :<= [C Eq _body]

-- specified but derived rules
C Eq Int :<= []
C Eq (Maybe _x) :<= [C Eq (DecP _name _vars _body)] % This should be C Eq <Maybe applied to _x so Just _ | Nothing>
C Eq (Foo _x) :<= [C Eq (DecP _name _vars _body)] % <Foo applied to _x so MkFoo Int String (Maybe _x)>
```

Now, when I ask `Eq (Foo Int)` what happens? Given this rule set:
1. `C Eq (Foo _x) :<= [C Eq (DecP _name _vars _body)] ` fires
2. `C Eq (DecP Foo _vars "MkFoo Int String (Maybe _x)") fires
3. Sum/Product rules kick in until
4. `C Eq String` is missing!!!
5. done

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
gnumonik: It's pattern matching substitution. It has its own rules. It's something extra you have* to graft onto the lambda calculus system. C.f. [here, esp 3.17.2](https://www.haskell.org/onlinereport/haskell2010/haskellch3.html#x8-580003.17.2)

* Well sorta, I read a paper once about a LC system where patterns aren't something extra grafted on, but I don't remember how it works.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
gnumonik: > This is very much unification.


> As I said, just by having ground terms on one side you can call it pattern matching, I'm still saying, this is AT LEAST covered by unification.

"X is Y" and "X is covered by Y" are two very different statements. You can of course implement pattern matching w/ a unification algorithm. But you do not _have to_. Unification is a more general / powerful / difficult to implement procedure than pattern matching. My references to the Haskell internals were just to demonstrate that you can implement a perfectly satisfactory pattern matching algorithm in a simpler way.

lambda-buffers-compiler/test/Test/TypeClassCheck.hs
gnumonik: ...I just remembered that `DecP` only exists because I was using patterns in the codegen and needed a pattern for the whole declaration. So yeah it's not really necessary, I guess. 

I suspect that there isn't something weird about what you're proposing too, though. Don't take the rest of this post as my final word on the matter b/c it might take me more than one try to put my finger on it. I'm just gonna think out loud: 

- `TyRef`s are just an identifier, right? If we were actually _evaluating_ anything here (we're not), we'd replace them with the body of the TyDef they refer to and then alpha conversion and beta reduction. They're morally equivalent to named functions in a lambda calculus system - i.e., they're necessary for user experience/sanity but from a certain point of view aren't necessary. 

- But they're also, kind of, not an identifier in that they're our version of `TyCon`. 

- I don't understand what `Res` is because we aren't actually evaluating. When you evaluate a function application in the lambda calculus you get a value (...me, just now realizing that's why it's called evaluation...). So `Res` would be the value that results from applying `F` to `a` for some (type) function `F` and some concrete type (`:: *`) `a`. I think what's confusing is me is that evaluating `Maybe Int` (i.e. `AppP (RefP _ (Name "Maybe")) (RefP _ (Name "Int))` gives us... `Maybe Int` (i.e. `AppP (RefP _ (Name "Maybe")) (RefP _ (Name "Int))`. 

- `A typeclass Eq is declaring that it can be applied to types of kind *.` That sounds right, but....

> 
    Sum/Prod/Nil
    TyApp - of kind *
    TyRef - of kind *
    TyAbs - of kind *
>

Is that really correct? I'm honestly not sure. 

I don't think `Sum/Prod/Nil` are really "types of kind *". It seems to me that they're the _values_ of _some concrete type_ which is of kind *.  If they're types then they definitely aren't of kind *, `Sum` has the kind `[[*]]` and `Prod` has the kind `[*]`.

gnumonik:
``` haskell
data Constraint = C Class Pat | Apply Pat Pat Pat | TyDef TyRef Ty
```

I have an overwhelmingly strong intuition that this isn't right. 

I can interpret `Apply` as a constraint if it's an equality constraint, e.g. `Apply F a ~ (F a)`. However the only thing we ever apply is a TyRef to a type w/ TyVar arguments, we never actually perform any reduction or normalization, and so the equality is trivial. I think in every possible case where we could use it, it's just `F a ~ F a` (or something that's a substitution instance of that).

gnumonik: To put the last point in a slightly different way: `solve` is the purely logical component of a larger system which has procedural components. `solve` emits subgoals that could not be determined on the basis of the rules that it has access to when we feed it a particular constraint. That's all it does / all it's meant to do. 

The `checkDerive` function that used to be in `Utils` contains the extra-logical components (resolving references & checking whether they're opaque, construcing the scope to feed to `solve`, etc). 

What you're suggesting is that we embed *everything* into the pure logical context. And that makes a _lot_ of sense if you're nourishing yourself on a diet of prolog because everything (more or less) in prolog only exists in a purely logical context of relations. I just think that embedding the resolution of `TyRef`s into rules doesn't make sense in Haskell; using `M.lookup` in a function that orchestrates invocations of `solve` strikes me as perfectly fine.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Pat.hs
```diff
@@ -19,10 +19,14 @@ cost of significantly more complex type signatures.
 -}
 
 data Pat
-  = {- extremely stupid, unfortunately necessary -}
+  = {- Name / ModuleName / Opaque / TyVarP are literal patterns (or ground terms)
+       because hey cannot contain any VarPs and therefore "have no holes".
+       Every TyDef or subcomponent thereof will be translated into a composite
+       pattern "without any holes". (Nil is also a literal/ground term, I guess) -}
```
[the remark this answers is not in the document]
gnumonik: They're literal patterns. 

```
--- yes i know this is a stupid way to write this 
isDrazen :: String -> Bool 
isDrazen str = case str of
  "Drazen" -> True     -- "Drazen" is a literal pattern here. It's not in an expression position so it can't be a (literal) expression 
   _ -> False          -- '_' is a wildcard pattern, which we don't have b/c they're not useful for us  
```

In our context they only match w/ identical literal patterns (in Haskell they match the identical literal expression). 

As an aside, it would be more "proper" to have an `Expr` type which is identical to `Pat` but w/o `VarP`. It would be _ideal_ to just use the ProtoCompat types, but we can't do that (without doing something ridiculous) because the proto AST isn't defined recursively (so there's no simple way to write `match`). 

I think a lot of your confusion/discomfort here comes from the fact that I used `Pat` to represent both proper patterns and "expressions" (patterns w/o `VarP` holes). If you think it would help I can go back and actually make an `Expr` type. (In the Haskell AST you can turn a pattern w/o `VarP` or `WildP` patterns into an expression simply by changing all of the LitP/ListP/etc values into LitE/ListE/etc values, so what I didn't isn't that wild really)

lambda-buffers-compiler/test/Test/TypeClassCheck.hs
gnumonik: > % IMPORTANT!!!: Just realized that this is unnecessary to be part of the Solver, could be, but rather we can construct these rules of each of the derive goals, which is what the user expects! AND WE DON'T NEED THE EXTRA Apply and TyDef 'utility' constraints

oh jfc i didn't scroll to the right and spent like an hour and a half today trying to figure out how to explain that to you lol

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
gnumonik: Just to be clear, when I point at Haskell internals here, it's because Haskell is _extensively studied and peer reviewed_ and we are _extremely unlikely to successfully reinvent all of the wheels_ that are relevant in this context.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
```diff
@@ -0,0 +1,131 @@
...
+  where
+    -- NOTE(@bladyjoker): The version w/ flip is more performant...
+    -- Given a Pat and a list of Classes, attempt to solve the constraints
+    -- constructed from the Pat and each Class
+    solveClassesFor :: Pat -> [Class] -> SolveM ()
+    solveClassesFor p = traverse_ (\cls -> solveM (C cls p))
```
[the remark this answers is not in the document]
gnumonik: You mean a global check for overlaps? That's, uh, doable I guess but has a complexity of `Olog(shit)`. We have to check whether any instance is a substitution of any other instance. 

If that's what you mean then let me know and I'll implement it. It's not particularly difficult.

lambda-buffers-compiler/test/Test/TypeClassCheck.hs
bladyjoker: lol sorry about that :D

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
```diff
@@ -0,0 +1,131 @@
...
+  [r] -> Right r
+  overlaps -> Left $ OverlappingMatches overlaps
+  where
+    matchPatAndClass :: Rule -> Bool
+    matchPatAndClass r =
+      ruleHeadClass r == c
+        && ruleHeadPat r
+        `matches` p
+
+type SolveM = ReaderT [Rule] (WriterT (S.Set Constraint) (Either Overlap))
```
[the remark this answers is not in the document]
gnumonik: I don't think `Map RuleHead Rule` is actually useful here? We're not doing straightforward lookups, we're filtering the List/Set/whatever of `Rule`s to find a match. A map doesn't really get us anything that we don't have w/ a list. 

From glancing at your protocompat PR it doesn't appear that `SolveError` is implemented ATM. That's fine, it might be easier to write the protos for all of the API errors after/inside of the next PR (since that's where the basic conditions for instance validity are)

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
gnumonik: Well, maybe. I missed something here: For all of the (other) instances that we assume to hold when checking a particular instance, it's probably safe to assume that their superclasses hold as well. I'm not 100% sure on this though and it needs more thought, but it's not really a problem *here* so much as it is a problem for #44

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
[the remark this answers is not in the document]
gnumonik: I think I'm going to leave things as they are for now. We can refactor this after #44 if you still think it's necessary. I'm 90% sure that the way things work in the next PR make it so that there's no real benefit to checking for overlaps at once (also duplicates just aren't possible b/c everything's a `Set` there)

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
```diff
@@ -0,0 +1,131 @@
...
+data MatchError
+  = NoMatch
+  | OverlappingMatches [Rule]
+
+-- for SolveM, since we catch NoMatch
+data Overlap = Overlap Constraint [Rule]
+  deriving stock (Show, Eq)
+
+selectMatchingInstance :: Pat -> Class -> [Rule] -> Either MatchError Rule
+selectMatchingInstance p c rs = case filter matchPatAndClass rs of
```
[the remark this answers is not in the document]
gnumonik: Oh wait I misunderstood what `RuleHead` was supposed to be below. Hm. 

I'm still not sure how this is an improvement tho. What you want is effectively `Map Constraint Rule` (I would say that `RuleHead ~ Pat`). If we're looking for the `Rule` that matches a given `Pat` for a given `Class` we can't just `lookup` using the `Class` and the `Pat`: It's an intrinsically `filter`-ey operation because we're looking for `RuleHead`s which _which are a generalization of the Pat / of which the Pat is a substitution instance_.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
```diff
@@ -0,0 +1,131 @@
...
+getSubs (RecP xs) (RecP xs') = getSubs xs xs'
+getSubs (SumP xs) (SumP xs') = getSubs xs xs'
+getSubs (AppP t1 t2) (AppP t1' t2') = getSubs t1 t1' <> getSubs t2 t2'
+getSubs (RefP n t) (RefP n' t') = getSubs n n' <> getSubs t t'
+getSubs (DecP a b c) (DecP a' b' c') = getSubs a a' <> getSubs b b' <> getSubs c c'
+getSubs _ _ = []
+
+-- NoMatch isn't fatal but OverlappingMatches is (i.e. we need to stop when we encounter it)
+data MatchError
+  = NoMatch
```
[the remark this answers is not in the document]
gnumonik: Yeah it is odd but the alternative is to use a `Maybe` where `Nothing` signals _success_ which is counterintuitive and not really what we want (and will probably do something weird in a `traverse`)

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
```diff
@@ -0,0 +1,131 @@
...
+  where
+    go :: [(Text, Pat)] -> Pat -> Pat
+    go subs tty =
+      let noflip p1 p2 = uncurry subV p2 p1
+       in foldl' noflip tty subs
+
+{- Given two patterns (which are hopefully structurally similar), gather a list of all substitutions
+   from the PatVars in the first argument to the concrete types (hopefully!) in the second argument
+-}
+getSubs :: Pat -> Pat -> [(Text, Pat)] -- should be a set, whatever
```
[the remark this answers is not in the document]
gnumonik: good point

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClass/Solve.hs
gnumonik: Assuming we reached an understanding on this important/relevant parts of this

bladyjoker (approved):

PR #48, opened 2023-02-14
Reworking the Compiler ProtoCompat (big one)

bladyjoker:
DONE
- APIfied all error messages
- Introduced Compiler API tests that use generators to create correct samples
- Refactored and reorganized for better structure


lambda-buffers-compiler/src/LambdaBuffers/Compiler/KindCheck/Inference.hs
```diff
@@ -96,10 +96,10 @@ infer ctx t = do
       mempty
         & context
           .~ M.fromList
-            [ (LocalRef "Σ", Type :->: Type :->: Type)
-            , (LocalRef "Π", Type :->: Type :->: Type)
-            , (LocalRef "𝟙", Type)
-            , (LocalRef "𝟘", Type)
+            [ (tyEither, Type :->: Type :->: Type)
```
cstml, line 99: `tySum` is consistent with `tyProd`
```suggestion
            [ (tySum, Type :->: Type :->: Type)
```

lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/Types.hs
```diff
@@ -68,6 +69,9 @@ data SourcePosition = SourcePosition {column :: Int, row :: Int}
   deriving stock (Show, Eq, Ord, Generic)
   deriving (Arbitrary) via GenericArbitrary SourcePosition
 
+defSourceInfo :: SourceInfo
+defSourceInfo = SourceInfo "" (SourcePosition 0 0) (SourcePosition 0 0)
+
```
cstml, line 74: I know community is against the use of it - but in this case it could be useful to do it via the [library and Class](https://hackage.haskell.org/package/data-default-0.7.1.1/docs/Data-Default.html)

lambda-buffers-compiler/test/Test/KindCheck.hs
```diff
@@ -1,18 +1,16 @@
...
 import Data.Bifunctor (Bifunctor (bimap))
-import Data.List.NonEmpty (NonEmpty ((:|)), cons)
 import Data.Text (Text)
 import LambdaBuffers.Compiler.KindCheck (
   check_,
   foldWithProduct,
   foldWithSum,
  )
-import LambdaBuffers.Compiler.KindCheck.Type (Type (App, Var))
+import LambdaBuffers.Compiler.KindCheck.Type (Type (App, Var), tyEither, tyProd, tyUnit, tyVoid)
```
cstml, line 10:
```suggestion
import LambdaBuffers.Compiler.KindCheck.Type (Type (App, Var), tySum, tyProd, tyUnit, tyVoid)
```

lambda-buffers-compiler/src/LambdaBuffers/Compiler/KindCheck/Inference.hs
cstml, line 99:
```suggestion
            [ (tySum, Type :->: Type :->: Type)
```

lambda-buffers-compiler/test/Test/KindCheck.hs
```diff
@@ -27,38 +25,16 @@ import Test.Tasty (TestTree, testGroup)
...
-  testCase "Two declarations of Maybe within the same module are caught." $
-    check_ compilerInput'doubleDeclaration
-      @?= Left (P.CompKindCheckError (P.MultipleTyDefError moduleName [tyDef'maybe, tyDef'maybe]))
-  where
-    moduleName = _ModuleName ["Module"]
-
-passingDoubleDeclaration :: TestTree
-passingDoubleDeclaration =
-  testCase "Two declarations of Maybe within different modules are fine." $
-    check_ compilerInput'doubleDeclarationDiffMod @?= Right ()
```
cstml, line 61: Have we added these somewhere else?

bladyjoker: In progress, 🙂

lambda-buffers-compiler/test/Test/KindCheck.hs
```diff
@@ -98,7 +74,7 @@ kcTestOrdering =
     genModuleIn2Layouts = do
       mods <- arbitrary
       shuffledMods <- shuffle mods
-      pure (P.CompilerInput mods, P.CompilerInput shuffledMods)
+      pure (_CompilerInput mods, _CompilerInput shuffledMods)
```
cstml, line 77: Because the `_CompilerInput` now creates a `Map` out of the shuffled list - the two `Maps` might not contain the same `TyDefs` - one map might contain one Module definition and the other another. 

If internally we're now dealing with `Maps` which are guaranteed by the `ProtoCompat` to error out prior if there are two instances of the same `Map` the above test can be removed - as long as it is covered by a test on the ProtoCompat parsing side.

bladyjoker: This was due to TyName being generated with the same name, so the indexing into Maps makes drastically different Modules. Will introduce this test case in the API tests, it's a good one.

lambda-buffers-compiler/test/Test/KindCheck.hs
bladyjoker: Will be part of a larger 'mutation testing' library.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/Types.hs
bladyjoker: I want to keep it ugly as this is a symptom that we should address proper. I'll add a todo above it.

lambda-buffers-compiler/app/LambdaBuffers/Compiler/Cli/Compile.hs
```diff
@@ -24,20 +21,19 @@ data CompileOpts = CompileOpts
...
-      NamingError ne -> print $ "Encountered a naming error " <> show ne
-      ProtoError pe -> print $ "Encountered a proto error " <> show pe
-    Right compIn' -> do
-      print @String "Successfully processed the CompilerInput"
-      let result = check compIn'
-      writeCompilerOutput (opts ^. output) (toProto result)
+  compInp <- readCompilerInput (opts ^. input)
+  case runCompiler compInp of
+    Left compErr -> do
+      print @String "Encountered errors during Compilation"
```
cstml, line 32:
```suggestion
      putStrLn "Encountered errors during Compilation:"
```

lambda-buffers-compiler/src/LambdaBuffers/Compiler/KindCheck.hs
```diff
@@ -36,51 +35,52 @@ import LambdaBuffers.Compiler.KindCheck.Inference (
...
 makeEffect ''GlobalCheck
 
 -- | Interactions that happen at the level of the
 data ModuleCheck a where -- Module
-  KCTypeDefinition :: ModName -> Context -> P.TyDef -> ModuleCheck Kind
-  KCClassInstance :: Context -> P.InstanceClause -> ModuleCheck ()
-  KCClass :: Context -> P.ClassDef -> ModuleCheck ()
+  KCTypeDefinition :: ModName -> Context -> PC.TyDef -> ModuleCheck Kind
+
+-- fixme(cstml & gnumonik): lets reach consensus on these - Note(1).
```
cstml, line 71:
```suggestion
-- NOTE(cstml & gnumonik): lets reach consensus on these - Note(1).
```

cstml (approved):

PR #38, opened 2023-02-08
Compiler proto small vocabulary fixes.

bladyjoker:
Hey @cstml can we do a quick run on the proto to do a quick cleanup.

PR #56, opened 2023-02-17
Compiler Proto Error Fixes

cstml:
Continuation from #38.


lambda-buffers-proto/compiler.proto
```diff
@@ -636,45 +636,56 @@ message ProtoParseError {
...
-  // Error referring to an unbound term. This usually means that the term was
-  // not defined.
-  message UnboundTermError {
-    TyName ty_name = 1;
-    VarName var_name = 2;
+  // Unbound variable ty_var detected in term ty_def.
+  message UnboundTyVarError {
+    TyDef ty_def = 1;
+    TyVar ty_var = 2;
+    ModuleName module_name = 3;
```
bladyjoker, line 643: Sorry mate for such a nit, but the module_name goes first as that's what we've being doing before. Would you mind putting it above?

lambda-buffers-proto/compiler.proto
```diff
@@ -636,45 +636,56 @@ message ProtoParseError {
...
-  // > ty_kind_1 with ty_kind_2.
+  // In ty_def an error has occurred when trying to unify kind ty_kind_lhs 
+  // with ty_kind_rhs.
   //
-  // FIXME(cstml): add source of constraint to the error such that user can see
+  // FIXME(cstml): Add source of constraint to the error such that user can see
   // where the constraint was generated - therefore where the error precisely
   // is.
   message ImpossibleUnificationError {
-    TyName ty_name = 1;
```
bladyjoker, line 655: Can we rename this into `UnificationError` assuming that it just means 2 things can't be unified right?

lambda-buffers-proto/compiler.proto
```diff
@@ -636,45 +636,56 @@ message ProtoParseError {
...
+  // error.
   message RecursiveKindError {
-    TyName ty_name = 1;
+    TyDef ty_def = 1;
+    ModuleName module_name = 2;
   }
 
-  // The inferred type differs from the type as defined.
+  // The actual_kind differs from the expected_kind.
   message InconsistentTypeError {
```
bladyjoker, line 675: Forgot if I asked, how's this different from the UnificationErr above?

lambda-buffers-proto/compiler.proto
```diff
@@ -636,45 +636,56 @@ message ProtoParseError {
...
 
-  // Inifinitely recursive term detected in definition ty_name.
+  // A cyclic term was encountered. Infinite terms like this are not acceptable,
+  // and we do not support them. We could not construct infinite "a = ty_name a"
+  // error.
   message RecursiveKindError {
-    TyName ty_name = 1;
+    TyDef ty_def = 1;
+    ModuleName module_name = 2;
   }
```
bladyjoker, line 672: If you could just elaborate on how this can be triggered by the API users that would be great. It says RecursiveKind but in the doc it says `cyclic term` which would mean `cyclic type` right?

lambda-buffers-proto/compiler.proto
cstml: This is to do with the inferred vs actual - so it happens after we have a Kind for the type which doesn't equal its declared type.

lambda-buffers-proto/compiler.proto
cstml: It means the same thing in this context.

lambda-buffers-proto/compiler.proto
cstml, line 643:
```suggestion
  message UnboundTyVarError {
    ModuleName module_name = 1;
    TyDef ty_def = 2;
    TyVar ty_var = 3;
```

lambda-buffers-compiler/test/Test/Utils/Module.hs
```diff
@@ -20,3 +20,27 @@ module'maybe = _Module (_ModuleName ["Module"]) [tyDef'maybe] mempty mempty
...
+  Should fail as Baz is a local undefined Ty Ref.
+-}
+module'undefinedLocalTyRef :: P.Module
+module'undefinedLocalTyRef = _Module (_ModuleName ["Module"]) [tyDef'undefinedLocalTyRef] mempty mempty
+
+{- | 1 Module containing
+  Foo b = Bar Module.Foreign.Baz b
+              ^^^^^^^^^^^^^^^^^^
+  Should fail as Baz is a foreign undefined Ty Ref.
+-}
```
bladyjoker, line 44: Nice!

lambda-buffers-proto/compiler.proto
bladyjoker: Hmm ok, my model would try to unify against a 'declared' and fail to do so if it doesn't match. That's why I ask. But it makes sense to have them separate as UnificationError might occur on any two terms.

bladyjoker (approved):
LGTM

PR #59, opened 2023-02-22
Removed InstanceClause, replaced with DeriveClause

gnumonik:
Small-ish change. It might be a good idea to leave `InstanceClause` in solely for communicating with the CodeGen component? Consider: 

```
data Foo a b = MkFoo a b 

derive Eq (Foo a b)
```

The typeclass-check stuff knows that the instance that needs to be derived is `instance (Eq a, Eq b) => Eq (Foo a b)`. And it is easy enough to determine this in the codegen - you just walk the SOP and collect free TyVars, since under "structural deriving" the context needs to contain `C a, C b, ...` for each TyVar (and only for each TyVar). Not sure what the best thing to do here is.

### DIFF
--- a/lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/FromProto.hs
+++ b/lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/FromProto.hs
@@ -553,23 +553,21 @@ instance IsMessage P.ClassDef PC.ClassDef where
       & P.documentation .~ doc
       & P.sourceInfo .~ toProto si
 
-instance IsMessage P.InstanceClause PC.InstanceClause where
+instance IsMessage P.DeriveClause PC.DeriveClause where
   fromProto ic = do
     si <- fromProto $ ic ^. P.sourceInfo
     cnm <- fromProto $ ic ^. P.classRef
-    csts <- traverse fromProto $ ic ^. P.constraints
     args <- traverse fromProto $ ic ^. P.args
     arg <- case args of
       [] -> throwInternalError "Zero instance arguments, but zero parameter type classes are not supported"
       [x] -> return x
       _ -> throwInternalError "Multiple instance arguments, but multi parameter type classes are not supported"
-    pure $ PC.InstanceClause cnm arg csts si
+    pure $ PC.DeriveClause cnm arg si
 
-  toProto (PC.InstanceClause cnm hd csts si) =
+  toProto (PC.DeriveClause cnm hd si) =
     defMessage
       & P.classRef .~ toProto cnm
       & P.args .~ pure (toProto hd)
-      & P.constraints .~ (toProto <$> csts)
       & P.sourceInfo .~ toProto si
 
 instance IsMessage P.Constraint PC.Constraint where
--- a/lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/Utils.hs
+++ b/lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/Utils.hs
@@ -43,7 +43,6 @@ import Data.Text (Text)
 import LambdaBuffers.Compiler.ProtoCompat.Types qualified as P (
   ClassDef,
   CompilerInput (CompilerInput),
-  Constraint (Constraint),
   DeriveClause (DeriveClause),
   Module,
   ModuleName,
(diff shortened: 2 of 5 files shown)


lambda-buffers-proto/compiler.proto
```diff
@@ -463,19 +463,16 @@ message TyClassRef {
...
 }
 
-/* Type class instances (rules)
+/* Type class instance deriving clauses
 
-Instance clauses enable users to specify 'semantic' rules for their types.
+Deriving clauses enable users to specify 'semantic' rules for their types.
 */
-message InstanceClause {
+message DeriveClause {
```
bladyjoker, line 470: wording nit; Initially I used "instance CLAUSE" to denote that it's a "clause" in a rule set (which is a set of clauses). I guess it's a pretty general term, can be applied in many settings but perhaps we could just call it `Derive` instead?

lambda-buffers-proto/compiler.proto
```diff
@@ -463,19 +463,16 @@ message TyClassRef {
...
   // Type class name.
   TyClassRef class_ref = 1;
-  // Instance (rule) arguments.
-  // Instance clause with no arguments is a trivial instance clause.
```
bladyjoker, line 474: Why did you remove this? I kind of liked the comment on this "edge" case.

lambda-buffers-proto/compiler.proto
```diff
@@ -508,7 +505,7 @@ message Module {
   // `ProtoParseError.MultipleClassDefError`.
   repeated ClassDef class_defs = 3;
   // Type class instance clauses.
-  repeated InstanceClause instances = 4;
+  repeated DeriveClause instances = 4;
```
bladyjoker, line 508: That would imo be `derive_goals` or smtn like that?

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/Utils.hs
```diff
@@ -276,29 +275,30 @@ buildClasses cis = foldM go M.empty (concat $ M.elems cis)
 
 type Instances = S.Set Instance
```
bladyjoker, line 277: You require a naming change here?

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/Utils.hs
```diff
@@ -276,29 +275,30 @@ buildClasses cis = foldM go M.empty (concat $ M.elems cis)
...
+    go :: S.Set Instance -> P.DeriveClause -> Either TypeClassError Instances
+    go acc (P.DeriveClause cn h si') = case ctable ^? ix cref of
       Nothing -> throwError $ UnknownClass cref si'
       Just cls -> do
         let p = tyToPat h
-        cs <- traverse goConstraint csts
-        let inst = C cls p :<= cs
+        let inst = C cls p :<= []
         checkInstance inst
         pure $ S.insert inst acc
```
bladyjoker, line 288: I think you need to adjust the naming at this point as it's getting confusing.

Imo you should internally use Rule? As `Instance` is sort of a Rule that a User can define, which is more restricted.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/ProtoCompat/FromProto.hs
```diff
@@ -553,23 +553,21 @@ instance IsMessage P.ClassDef PC.ClassDef where
...
+instance IsMessage P.DeriveClause PC.DeriveClause where
   fromProto ic = do
     si <- fromProto $ ic ^. P.sourceInfo
     cnm <- fromProto $ ic ^. P.classRef
-    csts <- traverse fromProto $ ic ^. P.constraints
     args <- traverse fromProto $ ic ^. P.args
     arg <- case args of
       [] -> throwInternalError "Zero instance arguments, but zero parameter type classes are not supported"
       [x] -> return x
       _ -> throwInternalError "Multiple instance arguments, but multi parameter type classes are not supported"
```
bladyjoker, line 564: Adjust the error messages here.

bladyjoker (approved):
I left a few comments, mostly about naming and adjusting the existing code to use a consistent naming scheme.

I'll approve this and let you merge at your convenience.

lambda-buffers-proto/compiler.proto
gnumonik: I think `DeriveClause` is OK. Maybe `DerivingClause`? I don't think "clause" has a single fixed technical meaning in this context.

PR #63, opened 2023-02-25
API Typeclass Errors 

gnumonik:
This is about the best I can do without attempting to rewrite the whole thing to use the ProtoCompat types (which I am fairly certain isn't possible to do in a sane manner). 

@bladyjoker You're not going to like the proto message types for the errors b/c they're basically just `(Text,SourceInfo)` but we might have to live with that unless you have a better idea of how to rewrite all of the typeclass machinery. You got your `SourceInfo` everywhere it matters, I'm not sure that I can meaningfully give you more than that. 

It should be possible to generate error messages which are frontend-agnostic enough to be meaningful independent of a particular frontend, however at the moment the errors render things into a Haskell-ish syntax. I can clean that up later, it's not important.

### DIFF
--- a/lambda-buffers-proto/compiler.proto
+++ b/lambda-buffers-proto/compiler.proto
@@ -633,71 +633,86 @@ message ProtoParseError {
   }
 }
 
-// Non-internal errors that can be thrown during
-// type class checking. Because these errors involve
-// TyClassCheck types, which cannot be converted back into
-// proto messages, the error message is constructed in the
-// TypeClassCheck modules in LambdaBuffers.Compiler
 message TypeClassError {
-  // Error which indicates that a class referred to in an instance
+  // A class referred to in an instance
   // or deriving clause does not exist. (Maybe this should be internal?)
   message UnknownClassError {
-    string msg = 1;
-    SourceInfo source_info = 2;
+    // The reference that could not be resolved
+    // NOTE: the class_ref contains valid/"real" source info
+    TyClassRef class_ref = 1;
   }
 
-  // Error which indicates that the TyDef referred to by a TyRef cannot be found
+  // The TyDef referred to by a TyRef cannot be found
   message LocalTyRefNotFoundError {
-    string msg = 1;
+    // NOTE: Has defaulted SourceInfo
+    TyName ty_name = 1;
+    // The module where the missing reference was searched for
+    ModuleName module_name = 2;
     // SourceInfo points to the clause that triggered the tyref lookup
-    SourceInfo source_info = 2;
+    SourceInfo source_info = 3;
   }
 
-  // Error which indicates a cycle of superclasses. SourceInfo
-  // isn't particularly meaningful here - we could include the SourceInfo
-  // for each class in the cycle, but given the large amount of work
-  // required to define a LB class, this should be more than adequate
+  // Detected a superclass cycle
   message SuperclassCycleError {
+    // TODO(@gnumonik): fix later if drazen is OK w/ the other ones,
+    // this one is not that important
     string msg  = 1;
   }
 
-  // Error that indicates a constraint cannot be solved with
+  // A constraint cannot be solved with
   // the in-scope set of instance rules.
-  message FailedToSolveConstraintError {
-    string msg = 1;
-    // SourceInfo points to the clause that triggered the subgoal failure
-    // (The failure could potentially be a structural rule, so pointing
-    // to the clause that triggered the exception is the best we can do)
-    SourceInfo source_info = 2;
+  message FailedToSolveConstraintsError {
+    // Name of the module where the failure to solve occurred
+    ModuleName module_name = 1;
+    // The constraints that could not be solved
+    // NOTE: ALL SourceInfos are defaulted here
+    repeated Constraint constraints = 2;
+    // The initial InstanceClause that triggered the failures
+    // NOTE: The outermost SourceInfo is correct for the clause,
+    //       but all inner SourceInfos are defaulted
+    InstanceClause instance = 3;
   }
 
-  // Error that indicates the body of a tydef is malformed in some way.
+  // The body of a tydef is malformed in some way.
   // Quasi-internal; indicates that something went wrong converting from
   // a TyDef to a Pat or Exp, and cannot be ameliorated by the user.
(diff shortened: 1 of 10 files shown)


lambda-buffers-proto/compiler.proto
```diff
@@ -633,6 +633,75 @@ message ProtoParseError {
...
+
+  // Error that indicates a constraint cannot be solved with
+  // the in-scope set of instance rules.
+  message FailedToSolveConstraintError {
+    string msg = 1;
+    // SourceInfo points to the clause that triggered the subgoal failure
+    // (The failure could potentially be a structural rule, so pointing
+    // to the clause that triggered the exception is the best we can do)
+    SourceInfo source_info = 2;
+  }
```
bladyjoker, line 672: I think this is the key error right? This is what we need if so:

```
// When solving a  `constraint`, a type class instance was found to be missing to solve the `missing_constraint`  in `module_name`.
message MissingInstanceError {
  ModuleName module_name = 1;
  Constraint constraint = 2;
  Constraint missing_constraint = 2;
}
```

lambda-buffers-proto/compiler.proto
```diff
@@ -633,6 +633,75 @@ message ProtoParseError {
...
+
+  // Error that indicates the body of a tydef is malformed in some way.
+  // Quasi-internal; indicates that something went wrong converting from
+  // a TyDef to a Pat or Exp, and cannot be ameliorated by the user.
+  // (I am putting it here because it has a meaningful SourceInfo)
```
bladyjoker, line 677: This sounds like an InternalError. Also, none of these comments should be in the proto. Think of these errors as implementation agnostic.

lambda-buffers-proto/compiler.proto
```diff
@@ -633,6 +633,75 @@ message ProtoParseError {
...
+
+  // Error that indicates an instance does not conform to one of the three
+  // basic conditions for instance validity. See the NOTE for `checkInstance`
+  // in LambdaBuffers.Compiler.TypeClassCheck.Utils. These conditions are all
+  // very straightforward and the message thoroughly explains what went wrong
+  message BadInstanceError {
+    string msg = 1;
+    // Points at the instance that triggered the violation
+    SourceInfo source_info = 2;
+  }
```
bladyjoker, line 692: No idea what this is. Make it InternalError or elaborate why this should be an API error.

bladyjoker (changes requested):
I generally think the set of errors are good, but there's still some work to do here. 

Note that API errors should be implementation agnostic, and we need to have these error messages descriptive such that they include the 'culprits' rather than a raw 'SourceInfo' which then needs explaining.

bladyjoker:
> @bladyjoker You're not going to like the proto message types for the errors b/c they're basically just `(Text,SourceInfo)` but we might have to live with that unless you have a better idea of how to rewrite all of the typeclass machinery. You got your `SourceInfo` everywhere it matters, I'm not sure that I can meaningfully give you more than that.

Where's the rush? You can definitely work on this and get it proper. Without looking at the implementation, I would say that instead of placing `SourceInfo` in your values threading it around, just put the `(ModuleName, Constraint)` or whatever the necessary information is. Would that be possible without rewriting the machinery? Concretely,

https://github.com/mlabs-haskell/lambda-buffers/blob/134ff00445aac6c09bfe63ecea433c3c708c3753/lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/Rules.hs#L31

Why not attach the PC.Constraint? Put it in a Maybe even so when you're going through structural rules you don't have access to it.

https://github.com/mlabs-haskell/lambda-buffers/blob/134ff00445aac6c09bfe63ecea433c3c708c3753/lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/Rules.hs#L18

Same here, just attach the information you know you'll need, like PC.ClassName.

I can't say exactly why it's so painful to implement this without looking at the code in more detail. But do take the time to get this right.

lambda-buffers-proto/compiler.proto
gnumonik:
```
Constraint missing_constraint = 2;
```

This is the problem (I'm assuming the first field 2 is the constraint that we're trying to solve and that the second is the actual failure case). 

The second field has a SourceInfo always because it's constructed from a `P.InstanceClause` or `P.DeriveClause` or whatever. 

The third field will always end up being one of the `assumptions`  (from `Validate.hs`): 

```
    assumptions =
      S.toList . S.fromList $
        S.toList inScopeInstances -- *A* the basic set of in-scope instances
          <> concatMap (mkStructuralRules . constraintClass) (c' : cs') -- *B* structural rules for all in scope classes
          <> S.toList (S.filter (/= i) . S.mapMonotonic unTag $ localInstances) -- *C* all instances in the current module that aren't the one we're trying to check
```

A and C can be tagged with a `PC.Constraint` since they're constructed from that originally. *B* can't be (and it is possible to end up with a failure involving Nil/Cons lists and variables "in the wrong place" for a `PC.Ty`). 

I can only think of two ways around this: 

1. Express the structural rules operationally instead of declaratively, which is bad because then you're definitely not getting a printout of them like you wanted. 
2. Tag everything with a `Maybe PC.Constraint` and have a meta-error type for... malformed errors... which seems incredibly stupid.

lambda-buffers-proto/compiler.proto
gnumonik: Did you look at LambdaBuffers.Compiler.TypeClassCheck.Utils? It explains what this is. There are 3 criteria an instance has to satisfy to ensure termination / soundness. I think I went over this in a meeting at some point.

lambda-buffers-proto/compiler.proto
gnumonik: Yeah what you're suggesting here is actually impossible. We also check that superclass constraints are satisfied as part of constraint solving, and those superclass constraints are always implicit / do not ever exist in the source. `missing_constraint` cannot contain a proto `Constraint` and expect the sourceinfo to mean anything at all.

gnumonik:
Honestly the implementation is kind of hideous but I don't have enough hours to go back and fix it all, unless you want to let me work into next week @bladyjoker

PR #64, opened 2023-02-27
Implemented superclass check with detailed reporting

bladyjoker:
DONE
- Superclass check
- Integrated typeclass checks into the Compiler pipeline
- Adjusted tests


lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck.hs
```diff
@@ -90,27 +47,29 @@ runDeriveCheck mn mb = mconcat <$> traverse go (S.toList $ mbInstances mb)
...
 
--- maybe use Control.Exception? Tho if we're not gonna catch it i guess this is fine
-validateTypeClasses :: P.CompilerInput -> IO (M.Map (P.InfoLess P.ModuleName) ModuleBuilder)
-validateTypeClasses ci = case validateTypeClasses' ci of
+runCheck :: PC.CompilerInput -> Maybe P.CompilerError
+runCheck ci = case Super.runCheck ci of
+  Left errs -> Just $ defMessage & P.tyClassCheckErrors .~ errs
+  Right () -> case runDeriveCheck' ci of
+    Left err -> Just $ defMessage & P.internalErrors .~ [defMessage & P.msg .~ ("TODO(bladyjoker): Use proper errors" <> (Text.pack . show $ err))]
+    Right _ -> Nothing
```
cstml, line 61: Don't use `Maybe` to return an error - it really doesn't behave in the appropriate way for this. 
Use Either and return unit.

bladyjoker: Yeah you're right. On it

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/SuperclassCycleCheck.hs
```diff
@@ -0,0 +1,98 @@
...
+        foldl'
+          ( \errs (mn, cd) ->
+              let errM = runReaderT (checkClassDef cd) (MkCheckRead mn (cd ^. #className) classDefs Set.empty [])
+               in case runExcept errM of
+                    Left err -> err : errs
+                    Right _ -> errs
+          )
+          []
+          [(m ^. #moduleName, cd) | m <- toList $ ci ^. #modules, cd <- toList $ m ^. #classDefs]
+   in if null allErrors then Right () else Left allErrors
```
cstml, line 50: This implementation is very hard to understand. As you're accumulating some errors and reporting if you found any refactor this to use a `WriterT [Err] Reader ...`  + traverse.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/SuperclassCycleCheck.hs
```diff
@@ -0,0 +1,98 @@
...
+
+checkClassDef :: MonadCheck m => PC.ClassDef -> m ()
+checkClassDef (PC.ClassDef cn _ sups _ _) = do
+  checkCycle cn
```
cstml, line 54: Make it clear you don't use the result - use `void` or `_ <-`.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/SuperclassCycleCheck.hs
bladyjoker: Good idea! On it!

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/SuperclassCycleCheck.hs
```diff
@@ -0,0 +1,98 @@
...
+    mn <- asks currentModuleName
+    currcn <- asks currentClassName
+    rtrace <- asks reportingTrace
+    throwError $
+      defMessage
+        & P.superclassCycleErr . P.moduleName .~ PC.toProto mn
+        & P.superclassCycleErr . P.className .~ PC.toProto currcn
+        & P.superclassCycleErr . P.cycledClassRefs .~ (PC.toProto <$> rtrace)
+
+throwUnboundTyClassRef :: MonadCheck m => PC.Constraint -> m ()
```
cstml, line 86: Does this have to be `()` in the return? The type is indicative that it is a useful result - make it explicit from the type that this is an error.
```
throwUnboundTyClassRef :: forall a {m} . MonadCheck m => PC.Constraint -> m a
```

cstml (changes requested):
Please address the above.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/SuperclassCycleCheck.hs
bladyjoker: Ok I tried but then realized that actually there's a reason I did it like this.

I can't just accumulate errors in Writer, because once the cycle error has been found it has to terminate the check for that particular classdef. So on each error I really do have to terminate the Error computation and throwError rather than tell.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/SuperclassCycleCheck.hs
bladyjoker: So with `throw` like actions you'd recommend to have `a` as the result generally?

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/SuperclassCycleCheck.hs
bladyjoker: The result is () so that makes it clear imo

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/SuperclassCycleCheck.hs
bladyjoker: I tried refactoring a bit to make it a bit more comprehensible.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/SuperclassCycleCheck.hs
```diff
@@ -0,0 +1,97 @@
...
+          , (cn, cd) <- Map.toList $ m ^. #classDefs
+          ]
+      -- Process each type class definitions under the appropriate reader context and collect errors
+      runClassDef (mn, cd) =
+        let errM = runReaderT (checkClassDef cd) (MkCheckRead mn (cd ^. #className) classDefs Set.empty [])
+         in case runExcept errM of
+              Left err -> [err]
+              Right _ -> []
+      allErrors = concat $ runClassDef <$> [(m ^. #moduleName, cd) | m <- toList $ ci ^. #modules, cd <- toList $ m ^. #classDefs]
+   in if null allErrors then Right () else Left allErrors
```
gnumonik, line 46: This is *extremely* hard for me to follow. Could you add type signatures to the `let` bindings, at least? I think that would help a lot w/ readability.

gnumonik (approved):
Aside from the not-really-a-readability-improvement-over-`concatMap` list comprehension this seems OK to me. I suspect the right way to do this would be to use a transformer stack with `[]` as the innermost monad, but it's somewhat atypical to use `do` notation w/ lists and perhaps that wouldn't actually be clearer.

lambda-buffers-compiler/src/LambdaBuffers/Compiler/TypeClassCheck/SuperclassCycleCheck.hs
bladyjoker: Done! I've introduced ProtoCompat.Indexing module that contains all the indexing we generally need so we don't reinvent the wheel. That should help.

bladyjoker:
@cstml I'm going to go ahead and merge this due to time constraints.

PR #82, opened 2023-04-13
Updating packages

bladyjoker:
DONE
- Refactored some bits using `flake-parts`
- Added and configured the HerculesCI `update-flake` action
- Upgraded all the packages
- Misc cleanup


aciceri (commented):
Some scattered notes:

* `flake-parts` lets you declare new options and then set values for them, this is a great pattern imo (the same used in NixOS), here you don't really do this, you just wrap everything inside `flake-parts.lib.mkFlake`. I understand that this could be not that easy since you already came up with `buildAbstraction` and `flakeAbstraction` without having `flake-parts` in mind.
* My suggestion is to keep in the `flake.nix` something like the following snippet. Split everything in different modules that you just import, doing this will force you to better structure your flake (exactly as you did for Hercules CI). Moreover other people might want to import some modules from your flake.
```nix
 inputs.flake-parts.lib.mkFlake { inherit inputs; } {
      systems = [ ... ];
      imports = [
        ...
      ];
    };
```
* You define a `pkgs` at line 27 when you have `pkgs` as `perSystem`'s argument. You probably do this because you need a `pkgs` "instance" with some overlays applied but this means that you would need to redo this in every module. I faced the same problem in the past and I came up with something like the following. It's also a nice example on how to use `flake-parts` idiomatically.
```nix
{
  inputs,
  lib,
  config,
  self,
  ...
}: {
  options.pkgsOverlays = let
    overlayType = lib.mkOptionType {
      name = "nixpkgs-overlay";
      description = "nixpkgs overlay";
      check = lib.isFunction;
      merge = lib.mergeOneOption;
    };
  in
    lib.mkOption {
      description = "Nixpkgs overlays to apply to `pkgs`";
      type = lib.types.listOf overlayType;
      default = with inputs; [ ];
    };

  config.perSystem = {
    system,
    lib,
    pkgs,
    ...
  }: {
    _module.args.pkgs =
      lib.foldl
      (legacyPackages: legacyPackages.extend)
      inputs.nixpkgs.legacyPackages.${system}
      config.overlays;
   };

   overlays = [ ... ];
};
```
* `pre-commit-hook.nix` provides its own `flake-parts` module, [here](https://github.com/cachix/pre-commit-hooks.nix/blob/master/template/flake.nix) an example of how to use it. Again, what you did is not wrong but you are not properly exploiting `flake-parts`.

bladyjoker:
> `pre-commit-hook.nix` provides its own `flake-parts` module, [here](https://github.com/cachix/pre-commit-hooks.nix/blob/master/template/flake.nix) an example of how to use it. Again, what you did is not wrong but you are not properly exploiting `flake-parts`.

Check out https://github.com/mlabs-haskell/lambda-buffers/pull/82/commits/31d38ac5e548cd6394484d19014a964b1e095e21 where I tried to do a module on pre-commit stuff.

bladyjoker:
> You define a `pkgs` at line 27 when you have `pkgs` as `perSystem`'s argument. You probably do this because you need a `pkgs` "instance" with some overlays applied but this means that you would need to redo this in every module. I faced the same problem in the past and I came up with something like the following. It's also a nice example on how to use `flake-parts` idiomatically.

Ok, that looks proper. I still don't know exactly how I would extract 'common' stuff into a new module, concretely:
1. All haskell modules should use the overlaid `pkgs`,
2. All modules should append the `pre-commit` `installationScript` to their `shellHook`
3. All haskell modules should use a specific `compiler` and `index-state`

I imagine I need to create a small module with default options and then instantiate such a module.

bladyjoker:
If you look at all the `build.nix` files in haskell directories you'll see:

```nix
{ pkgs
, haskell-nix
, mlabs-tooling
, compiler-nix-name
, index-state
, compilerHsPb
, commonTools
, shellHook
}:
let
  inherit pkgs;
  project = {
    src = ./.;

    name = "lambda-buffers-compiler";

    inherit compiler-nix-name index-state;

    extraHackage = [
      "${compilerHsPb}"
    ];

    modules = [
      (_: {
        packages = {
          allComponent.doHoogle = true;
          allComponent.doHaddock = true;

          # Enable strict compilation
          lambda-buffers-compiler.configureFlags = [ "-f-dev" ];
        };
      })
    ];

    shell = {

      withHoogle = true;

      exactDeps = true;

      nativeBuildInputs = [ pkgs.swiPrologWithGui ] ++ builtins.attrValues commonTools;

      tools = {
        cabal = { };
        haskell-language-server = { };
      };

      shellHook = ''
        export LC_CTYPE=C.UTF-8
        export LC_ALL=C.UTF-8
        export LANG=C.UTF-8
        ${shellHook}
      '';
    };
  };
in
{
  hsNixProj = haskell-nix.cabalProject' [
    mlabs-tooling.lib.mkHackageMod
    project
  ];
}
```

which is tedious to maintain.

bladyjoker:
Thanks @aciceri

PR #102, opened 2023-08-15
Flake update

bladyjoker:


bladyjoker:
@szg251 could you please take a look at this. We have some packages that come from CTL and some other places that are problematic and mess with our automatic flake updates. Would you mind playing with this a bit?

Issue #103, opened 2023-10-02
Create a Rust runtime for LB Prelude


szg251:
- [x] json serialisers and deserialisers
- [x] roundtrip tests
- [x] nix integration

PR #105, opened 2023-10-02
Create a Rust runtime for LB Prelude

szg251:
Closes: #103

### nix integration
I used nix-cargo-integration (which builds on dream2nix). This hides a lot of complexity with reasonable default configurations. It also uses flake-parts, so making the `build.nix` a flake module seemed like a natural choice, but it diverges from the current status quo, so there's a separate issue to bring everything to the same conventions.

### Types
Integer: num_bigint is used in cardano_serialisation_lib
ByteString: `&[u8]` or `Vec<u8>` are the obvious choices, I'm not sure if there's too much difference from the end-user standpoint.

### Json serialisers
I followed the same principles that we used in Hs and Ps packages: `to_json` instances serialise to a `Value` (intermediary JSON representation). However I'm not sure, if this is the most convenient way from the end-user perspective, we will have to experiment with this later.

Another question mark is error messages: I'm still not familiar with the messages `serde` returns, so I basically copied the error messages from the Haskell implementation. To make this library feel more natural, I think these should be adjusted in the future.

### Property testing
There are two popular options in Rust for property testing: `quickcheck` and `proptest`.
The former is a simple library, mostly following the Haskell implementation and design principles. `proptest` is a more robust (and a little bit slower) tool, with a lot of nice features.
I decided to go with the latter one, as it allowed me to write simply implement BigInt generators. Also, I simply liked the API of the library.
One area which is still not clear to me is `proptest-regressions`. When a proptest fails, the input values are stored in the project directory, and used in later tests. I decided to put this into `.gitignore`, I'm not exactly sure if this is working exactly as I desired (e.g. these tests logs remain even if I change the tests). 

## Links to related issues
- #109
- #110


flake.nix
```diff
@@ -20,6 +21,7 @@
         (import ./hercules-ci.nix)
         (import ./pre-commit.nix)
         (import ./docs/build.nix)
+        ./runtimes/rust/lbr-prelude/build.nix
```
bladyjoker, line 24: Wait, this works? If it works, the import statements above could also just be file literals? Please try pretty please!

runtimes/rust/lbr-prelude/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "lbr-prelude"
+version = "0.1.0"
+edition = "2021"
```
bladyjoker, line 4: 2023 maybe? Just saying :)

runtimes/rust/lbr-prelude/build.nix
```diff
@@ -0,0 +1,16 @@
+{ inputs, ... }: {
+  imports = [ inputs.nci.flakeModule ];
+  perSystem = { pkgs, system, inputs', config, ... }:
+    let crateName = "lbr-prelude";
+    in {
+      nci.projects.${crateName}.path = ./.;
+      nci.crates.${crateName} = { };
+
+      devShells."dev-${crateName}-rust" = config.nci.outputs.${crateName}.devShell.overrideAttrs (old: {
+        packages = (old.packages or [ ]) ++ [ pkgs.rust-analyzer ];
```
bladyjoker, line 10: Is this the LSP?

runtimes/rust/lbr-prelude/src/json.rs
```diff
@@ -0,0 +1,327 @@
...
+
+impl Json for BigInt {
+    fn to_json(&self) -> Result<Value, Error> {
+        serde_json::to_value(self)
+    }
+
+    fn from_json(value: Value) -> Result<Self, Error> {
+        serde_json::from_value(value)
+    }
+}
```
bladyjoker, line 27: Can you put a list mapping LB Prelude opaque types into Rust types, and let's have a brief look around that we indeed have chosen what rustaceans would consider proper

runtimes/rust/lbr-prelude/src/json.rs
```diff
@@ -0,0 +1,327 @@
...
+                .into_iter()
+                .collect::<Result<Vec<T>, Error>>(),
+            _ => error(&format!("Expected a JSON Array but got {:?}", value)),
+        }
+    }
+}
+
+impl<T> Json for HashSet<T>
+where
+    T: Json + Eq + Hash,
```
bladyjoker, line 174: Addition of `Hash` here probably seems benign, but that means that during code generation, we have to derive Hash for all LB types we map into Rust.

Feel free to add a `NOTE(syg251): Must add Hash derivation to type def printing`

runtimes/rust/lbr-prelude/src/json.rs
```diff
@@ -0,0 +1,327 @@
...
+            },
+            _ => error("Expected a JSON Array with 2 fields"),
+        }
+    }
+}
+
+/// Create a JSON serialisation error with a custom message
+fn error<T>(msg: &str) -> Result<T, Error> {
+    Err(ser::Error::custom(msg))
+}
```
bladyjoker, line 293: I know I didn't do that in the Hs/Purs counterparts, but I like the pattern of having an `errors.rs` module where the errors used by all (sub)modules. There we can enumerate all the errors in some convenient form and have a single place to control how it;s printed.

For example:
```rust
#[derive(thiserror::Error, Debug)]
enum Error {
  #[error("Expected a JSON type: {1}, but got a JSON type: {0}")]
  UnexpectedJsonType {got: JsonType, wanted: JsonType},
  #[error("Expected a field name in a JSON Object: {1}, but got fields named {0}")]  
  UnexpectedFieldName {got: Vec<String>, wanted: String},
  #[error("Expected a JSON Array with {1} elements, but got {0} elements")]  
  UnexpectedArrayLength {got: int, wanted: int},
  #[error("Some internal error happened: {0}")]
  InternalError(String),
}
```

bladyjoker: I want to introduce that in our Hs/Purs base, Rust folks have good simple conventions that improve everyone's lives so why not cross polinate.

bladyjoker (commented):
Great freaking job! Left some comments, let me know wdyt.

runtimes/rust/lbr-prelude/src/json.rs
```diff
@@ -0,0 +1,322 @@
...
+use std::hash::Hash;
+
+/// Trait that lbf-prelude::json class maps to
+pub trait Json {
+    fn to_json(&self) -> Result<Value, Error>;
+
+    fn from_json(value: Value) -> Result<Self, Error>
+    where
+        Self: Sized;
+}
```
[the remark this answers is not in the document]
szg251: My reasoning for this was that Rust always uses the package name in imports, so I thought we don't need all the namespaces.
The crate name is `lbr-prelude`, so it would become `lbr-prelude::runtime::prelude::json::*`? Which already implies lambda buffers runtime prelude.
I generally agree with the notion of consistency, but I'm a bit indecisive here, if it's necessary here.

flake.nix
szg251: Tried it, we can remove the imports

runtimes/rust/lbr-prelude/Cargo.toml
szg251: I just checked, the latest edition is 2021 (if this document is honest: https://doc.rust-lang.org/edition-guide/editions/index.html )

runtimes/rust/lbr-prelude/src/json.rs
szg251: > Fyi, I'm using this notation for comments that can be collected and analyzed by 3rd party tooling

That's nice, I wanted to introduce something like that in another project

If we use HashSet (which seems to be a reasonable choice), we must have the Hash trait implemented for the key. But now that you mention it, I think we could provide a more generic implementation for anything implementing the Iterator trait. This would tick out Vec, HashSet and HashMap, and a lot more. I'll play around with this for a bit...

runtimes/rust/lbr-prelude/src/json.rs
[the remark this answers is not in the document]
bladyjoker: I agree, let's keep it as it is, `lbr-prelude::Json` is sufficient!

runtimes/rust/lbr-prelude/Cargo.toml
bladyjoker: omg I thought this is just for our package somehow, ignore me

runtimes/rust/lbr-prelude/src/json.rs
szg251: Sounds like a good idea.

runtimes/rust/lbr-prelude/src/json.rs
```diff
@@ -0,0 +1,327 @@
...
+    }
+}
+
+/// Create a JSON serialisation error with a custom message
+fn error<T>(msg: &str) -> Result<T, Error> {
+    Err(ser::Error::custom(msg))
+}
+
+/// Construct a JSON Value from a sum type
+/// We always encode sum types into a `{"name": string, "fields": any[]}` format in JSON
```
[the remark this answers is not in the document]
szg251: I love these undocumented traps.

runtimes/rust/lbr-prelude/src/json.rs
```diff
@@ -0,0 +1,327 @@
...
+                }
+            }
+            _ => error(&format!("Expected a JSON Array but got {:?}", value)),
+        }
+    }
+}
+
+impl<K, V> Json for HashMap<K, V>
+where
+    K: Json + Eq + Hash,
```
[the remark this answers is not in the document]
szg251: hmm, but in that case we would have to convert to a Value first, which feels like a bit of an unnecessary overhead.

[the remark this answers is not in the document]
bladyjoker: Looks to me like key.to_json() is already doing that no?

runtimes/rust/lbr-prelude/src/json.rs
```diff
@@ -0,0 +1,327 @@
+use data_encoding::BASE64;
+use num_bigint::BigInt;
+use serde::ser;
+use serde_json::{self, Error, Value};
+use std::collections::{HashMap, HashSet};
+use std::hash::Hash;
+
+/// Trait that lbf-prelude::json class maps to
+pub trait Json {
+    fn to_json(&self) -> Result<Value, Error>;
```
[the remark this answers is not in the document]
szg251: I also found it strange but `serde_json` returns a Result in it's [`to_value` ](https://docs.rs/serde_json/1.0.107/serde_json/fn.to_value.html#) method, so I had to follow that.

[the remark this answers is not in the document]
szg251: The explanation is this: 
> This conversion can fail if T’s implementation of Serialize decides to fail, or if T contains a map with non-string keys.

pre-commit.nix
```diff
@@ -34,6 +34,8 @@
             markdownlint.enable = true;
             dhall-format.enable = true;
             purty.enable = true;
+            # TODO: Enable rustfmt
+
```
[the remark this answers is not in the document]
szg251: Addressed in #113

runtimes/rust/lbr-prelude/build.nix
szg251: Yes (this config is now overwritten in #113)

runtimes/rust/lbr-prelude/src/json.rs
```diff
@@ -0,0 +1,327 @@
...
+    V: Json,
+{
+    fn to_json(&self) -> Result<Value, Error> {
+        let values = self
+            .iter()
+            .map(|(key, val)| Ok(Value::Array(vec![key.to_json()?, val.to_json()?])))
+            .into_iter()
+            .collect::<Result<Vec<Value>, Error>>()?;
+
+        Ok(Value::Array(values))
```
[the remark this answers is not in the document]
szg251: I couldn't find any explicit ordering, however both PS and Haskell are using binary trees, so keys are ordered. We could use `BTreeMap` on the Rust side, to keep things consistent.

runtimes/rust/lbr-prelude/src/json.rs
szg251: I'm thinking about using BTreeMap instead of a HashSet as a default set implementation, but I think it's not the common choice.
Using BTreeMap would have these two benefits:
- same constraints apply as in HS and PS (Ord instead of Hash)
- json instanced are "automatically" come out ordered
What do you think?

runtimes/rust/lbr-prelude/src/json.rs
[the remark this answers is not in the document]
szg251: Same idea applies here: BTreeMap instead of HashMap

runtimes/rust/lbr-prelude/src/json.rs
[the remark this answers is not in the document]
szg251: a Hash trait implementation is required for the key, when we build the HashMap of the deserialized Map, so the original key must be hashable.
This would also change to an `Ord` instead of `Hash` if we choose `BTreeMap`.

runtimes/rust/lbr-prelude/src/json.rs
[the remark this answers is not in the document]
szg251: With the last commit, I decided to eliminate `to_value` everywhere.
The last place where encoding could fail is, but in theory that should be impossible. Should we just throw an error instead?

runtimes/rust/lbr-prelude/src/json.rs
bladyjoker: cp from the chat

You decide, it's ok to do whatever and adjust after if needed. I think Prelude.Map encoding needs to be documented. I 'think' that by accident the keys are ordered, and I would say "let's make that part of the LB.Json encoding standard for Map".
So {b: 1, a: 2} dictionary is serialized as [["a", 2], ["b", 1]].

runtimes/rust/lbr-prelude/src/json.rs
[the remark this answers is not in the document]
bladyjoker: Leaving this for you to decide, we can easily adjust if needed.

bladyjoker (approved):
This is lovely!

Let me know if you need help syncing this with main.

PR #113, opened 2023-10-12
Custom pre-commit hooks for our monorepo setup

szg251:
- Configuring nix for rust prelude
- Remove rust test from checks
- Naive implementation of JSON serialisers/deserialisers in Rust
- Removing cargo workspaces
- Move rust specific gitignores to package directory
- Implement JSON roundtrip tests
- Experimenting with flakes modules
- Remove crates.nix
- Delete proptest-regressions file
- Add documentation to exposed functions
- Remove 'import' statement for flake module imports
- Add rustfmt and clippy


.envrc
```diff
@@ -1 +1,2 @@
+nix_direnv_watch_file ./pre-commit.nix
```
bladyjoker, line 1: Was is das? Just curious :)

szg251: I think I mentioned this in our session, we can set direnv to reload automatically, when some files are changed (this takes multiple arguments)

pre-commit.nix
```diff
@@ -1,14 +1,15 @@
-{ inputs, lib, ... }: {
+{ inputs, ... }: {
   imports = [
-    inputs.pre-commit-hooks.flakeModule # Adds perSystem.pre-commit options
+    ./extras/pre-commit-hooks-extra.nix
```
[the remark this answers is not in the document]
szg251: I think we covered this as well..

bladyjoker (approved):
I think you got this :+1: 

I'm happy about `deadnix` :D

extras/pre-commit-hooks-extra.nix
```diff
@@ -0,0 +1,20 @@
...
+  ];
+  perSystem = { pkgs, ... }:
+    let
+      rustfmt = pkgs.rust-bin.stable.latest.rustfmt;
+    in
+    {
+      pre-commit.settings.hooks = {
+        rustfmt-monorepo =
+          {
+            name = "rustfmt";
```
szg251, line 13: Ended up with a ridiculously simple configuration, now that clippy is out of the picture

PR #126, opened 2023-10-25
Updated `docs/haskell.md` and `docs/purescript.md` to reflect the updated frontend tools

jaredponn:
A short PR to update the documentation to use the tools in #121. This also:
- Changed the example to use lambda buffers `prelude` and hence use the opaque `Char` type (it won't compile otherwise without it)

### DIFF
--- a/docs/examples/Document.lbf
+++ b/docs/examples/Document.lbf
@@ -1,7 +1,7 @@
 module Document
 
--- We need the opaque Char type
-import Prelude (Char)
+-- Importing types
+import Prelude (Text, List, Set, Bytes)
 
 -- Author
 sum Author = Ivan | Jovan | Savo
@@ -12,7 +12,7 @@ sum Reviewer = Bob | Alice
 -- Document
 record Document a = {
   author : Author,
-  reviewers : List Reviewer,
+  reviewers : Set Reviewer,
   content : Chapter a
  }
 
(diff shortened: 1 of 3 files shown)


docs/examples/Document.lbf
```diff
@@ -1,5 +1,8 @@
 module Document
 
+-- We need the opaque Char type
+import Prelude (Char)
+
```
bladyjoker, line 5: Yeah, was trying to document this whilst using no `opaques`. Not sure why x)

docs/examples/Document.lbf
```diff
@@ -35,14 +38,5 @@ prod RichDocument = (Document RichContent)
 -- ## We need a list type
 sum List a = Nil | Cons a (List a)
```
bladyjoker, line 40: If this is a Prelude documentation, then we can use Prelude.List opaque as well?

docs/examples/Document.lbf
```diff
@@ -35,14 +38,5 @@ prod RichDocument = (Document RichContent)
...
-sum Letter = A | B | C
-
-sum Number = Num0 | Num1 | Num2
-
-sum Punctuation = Dot | Question
-
 -- ## String
-prod String = (List Char)
\ No newline at end of file
+prod String = (List Char)
```
bladyjoker, line 42: Let's use Prelude.Text opaque?

bladyjoker (changes requested):
You're right, let's use Prelude for this:

That would change the schema for Document a bit...

```lbf
module Document

-- We need the opaque Char type
import Prelude (Text, List, Set, Bytes)

-- Author
sum Author = Ivan | Jovan | Savo

-- Reviewer
sum Reviewer = Bob | Alice

-- Document
record Document a = {
  author : Author,
  reviewers : Set Reviewer,
  content : Chapter a
 }

-- Chapter
record Chapter a = {
  content : a,
  subChapters : List (Chapter a)
 }

-- Some actual content
sum RichContent = Image Bytes | Gif Bytes | Text Text

-- Rich document
prod RichDocument = (Document RichContent)
```

bladyjoker (approved):
Thank you!

PR #130, opened 2023-10-27
Optimized the `Eq` instance generation by removing the useless `True` at the start

jaredponn:
This optimizes the generated `Eq` instances.

For example,
```
module TEST

import Prelude

record A = { int : Integer }
derive Eq A
```
would previously generate code like
```
instance Prelude.Eq A where
  (==) = (\x0 -> (\x1 -> (Prelude.&&) (Prelude.True) ((Prelude.==) (a'int x0) (a'int x1)) ) )
                                   -- ^~~~~~~~~~~~~ this is not necessary
``` 

This PR changes the generated code to be as follows
```
instance Prelude.Eq A where
  (==) = (\x0 -> (\x1 -> (Prelude.==) (a'int x0) (a'int x1) ) )
```


bladyjoker (approved):
Thanks!

PR #135, opened 2023-11-09
Upgrading to CTL latest

### LINKED ISSUE #106: Sync with CTL
https://github.com/Plutonomicon/cardano-transaction-lib/issues/1472

Is changing to using a different BigInt representation instead of Data.BigInt that we code gen into: https://github.com/mlabs-haskell/lambda-buffers/blob/be0d0c8a5a925bd07243fcae58dfc46c0ce180e3/lambda-buffers-codegen/data/purescript.json#L45

Consolidate!

bladyjoker:
Upgrades to https://github.com/Plutonomicon/cardano-transaction-lib/pull/1551 which is still to be merged in develop but we need it anyway.

In each Purs project:
1. Update packages.dhall and `spago2nix generate`
2. Update package.json and `npm i`
3. Change the package.json to `"type" : ":module"`
4. Added `esbuild/bundle.js` file

Closes #106


szg251 (approved):
Looks good. I wonder, if we could (or should) abstract the `bundle.js` somewhere.

bladyjoker:
> Looks good. I wonder, if we could (or should) abstract the `bundle.js` somewhere.

God damn tried, but node doesn't accept a file as a ES module if it's not in the same directory as smtn else I dunno, this whole stack is anxiety provoking

PR #138, opened 2023-11-10
Demo: Expose lbf-nix for users

bladyjoker:
DONE:
- [x] Refactore the haskell build.nix-es to use the `extras/flake-haskell.nix`
- [x] Completely remove our dependency on mlabs-haskell/mlabs-tooling.nix and use Chap and Plutus direclty (better control there)
- [x] Made lbf-nix into a reusable form via flake-parts module.


bladyjoker:
Unsanctioned!

PR #142, opened 2023-11-13
Enable MacOS builds and CI checks

szg251:
For some reason `x86_64-darwin` was removed from the supported systems.


szg251:
@bladyjoker I actually found out that we can run darwin on Hercules, I don't know why I didn't notice earlier. This led to the discovery that some packages are not building on Mac, so I plan to have a look at this.
But I guess it's not a high priority right now, so maybe I can return to this after the Rust milestone is done.

bladyjoker:
Oh sorry, why did I close this then! reopening

bladyjoker (approved):

PR #149, opened 2023-11-17
Add simple Emacs mode

t4ccer:
Simple mode derived from Haskell mode to not write custom parsers in this rather early stage. Indentation is a bit janky but you can hit tab couple times to get what you mean. Syntax highlighting works just fine.

Can be used with `straight.el` or in any other way you use to import packages from git
```elisp
(use-package lambda-buffers-mode
  :straight
  ( :host github
    :repo "mlabs-haskell/lambda-buffers"
    :branch "t4/add-simple-emacs" ;; remove after merge
    :files ("extras/editor/emacs/lambda-buffers-mode.el")))
```


bladyjoker (commented):
Woohooo!!! Would you mind pleeease putting this into `extras`? So extras/editor/emacs/lambda-buffers-mode.el

bladyjoker:
There's no lisp pre-commit-hooks ofc, I'll add an issue for adding smtn like that to our pre-commit-hooks https://github.com/gonewest818/elisp-lint

extras/editor/emacs/lambda-buffers-mode.el
```diff
@@ -0,0 +1,16 @@
+;;; lambda-buffers-mode.el --- SUMMARY -*- lexical-binding: t -*-
+
+(require 'haskell-mode)
+
+(define-derived-mode lambda-buffers-mode haskell-mode "lambda-buffers"
+  "haskell-mode pretending to be lambda-buffers-mode"
+  (haskell-indent-mode)
+  (font-lock-add-keywords
+   nil
+   '(("\\<\\(record\\|sum\\|derive\\|opaque\\|prod\\)\\>" . 'font-lock-keyword-face)))
```
bladyjoker, line 10: We also have `instance` keyword, should we add it here?

Instance definitions are a declaration of 'this semantics is supported by the implementation' so for example, we have a `Maybe` opaque type in Prelude schema and say `instance Eq (Maybe a) :- Eq a`. The syntax is Prolog syntax :D because it was easier to parse

bladyjoker (approved):

PR #150, opened 2023-11-18
Typescript runtime

jaredponn:
Prelude TODOs:
- [x] Prelude runtime (mostly)
- [x] Prelude runtime using a sensible Map / Set to something else that doesn't use strange (to Haskell  programmers) notions of equality.
- [x] Prelude runtime tests (mostly)
- [x] Tsdocs
- [x] Map / set tests

Plutus TODOs
- [x] Plutus runtime
- [x] Proper newtypes around select bytestring wrappers.
- [x] Plutus runtime tests
- [x] Tsdocs

### DIFF
--- a/lambda-buffers-codegen/src/LambdaBuffers/Codegen/Haskell/Print/Derive.hs
+++ b/lambda-buffers-codegen/src/LambdaBuffers/Codegen/Haskell/Print/Derive.hs
@@ -8,7 +8,6 @@ import Data.Set (Set)
 import LambdaBuffers.Codegen.Haskell.Print (MonadPrint)
 import LambdaBuffers.Codegen.Haskell.Print.LamVal (printValueE)
 import LambdaBuffers.Codegen.Haskell.Print.Syntax qualified as H
-import LambdaBuffers.Codegen.LamVal qualified as LV
 import LambdaBuffers.Codegen.LamVal.Eq (deriveEqImpl)
 import LambdaBuffers.Codegen.LamVal.Json (deriveFromJsonImpl, deriveToJsonImpl)
 import LambdaBuffers.Codegen.LamVal.MonadPrint qualified as LV
@@ -55,14 +54,15 @@ hsClassImplPrinters =
 eqClassMethodName :: H.ValueName
 eqClassMethodName = H.MkValueName "=="
 
-lvEqBuiltinsBase :: Map LV.ValueName (H.CabalPackageName, H.ModuleName, H.ValueName)
-lvEqBuiltinsBase =
-  Map.fromList
-    [ ("eq", (H.MkCabalPackageName "base", H.MkModuleName "Prelude", H.MkValueName "=="))
-    , ("and", (H.MkCabalPackageName "base", H.MkModuleName "Prelude", H.MkValueName "&&"))
-    , ("true", (H.MkCabalPackageName "base", H.MkModuleName "Prelude", H.MkValueName "True"))
-    , ("false", (H.MkCabalPackageName "base", H.MkModuleName "Prelude", H.MkValueName "False"))
-    ]
+lvEqBuiltinsBase :: LV.PrintRead (H.CabalPackageName, H.ModuleName, H.ValueName)
+lvEqBuiltinsBase = LV.MkPrintRead $ \(_ty, refName) ->
+  Map.lookup refName $
+    Map.fromList
+      [ ("eq", (H.MkCabalPackageName "base", H.MkModuleName "Prelude", H.MkValueName "=="))
+      , ("and", (H.MkCabalPackageName "base", H.MkModuleName "Prelude", H.MkValueName "&&"))
+      , ("true", (H.MkCabalPackageName "base", H.MkModuleName "Prelude", H.MkValueName "True"))
+      , ("false", (H.MkCabalPackageName "base", H.MkModuleName "Prelude", H.MkValueName "False"))
+      ]
 
 printDeriveEqBase :: MonadPrint m => PC.ModuleName -> PC.TyDefs -> (Doc ann -> Doc ann) -> PC.Ty -> m (Doc ann)
 printDeriveEqBase mn iTyDefs mkInstanceDoc ty = do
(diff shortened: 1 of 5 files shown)


extras/flake-typescript.nix
```diff
@@ -0,0 +1,186 @@
...
+
+in
+{
+  devShells = {
+    # Note: when using the devshell, when in the project root directory, type
+    # ```
+    # ln -s "$NODE_PATH" node_modules
+    # ```
+    # so when `node` executes, it will use the dependencies gathered together
+    # by nix
```
bladyjoker, line 173: Why? Why don't you just link these? Also this won't work right if NODE_PATH contains more than one paths separated by a :

extras/pre-commit-hooks-extra.nix
```diff
@@ -15,6 +16,27 @@
             entry = "${rustfmt}/bin/rustfmt --color always";
             files = "\\.rs$";
           };
+
+
+        # TODO(jaredponn): I'm fairly certain later versions of flake-parts
+        # includes `denofmt` and `denolint` so we don't have to reimplement
+        # this ourselves
```
bladyjoker, line 23: You mean pre-commit-hooks.nix? `nix flake lock --update-input <>` whatever needs to be updated.

runtimes/typescript/lbr-plutus/build.nix
```diff
@@ -0,0 +1,16 @@
...
+        config.overlayAttrs.extras.typescriptFlake {
+          name = "lbr-plutus-typescript";
+          src = ./.;
+          inherit system;
+          dependencies = [ config.packages."lbr-prelude-typescript-tarball" ];
+        };
+    in
+    {
+      inherit (typescriptFlake) packages checks devShells;
+    };
```
bladyjoker, line 15: very nice!

runtimes/typescript/lbr-plutus/src/LambdaBuffers/Runtime/PlutusData.ts
```diff
@@ -0,0 +1,245 @@
...
+      return LbPrelude.eqInteger.neq(l.fields, r.fields);
+    } else {
+      return true;
+    }
+  },
+};
+
+export const toDataPlutusData: ToData<PlutusData> = {
+  toData: (arg) => arg,
+};
```
bladyjoker, line 83: I love what you did with type classes

runtimes/typescript/lbr-prelude/src/LambdaBuffers/Runtime/Prelude.ts
```diff
@@ -0,0 +1,700 @@
+import type { Json } from "./Json.js";
+import { JsonError, Scientific } from "./Json.js";
+import * as LbJson from "./Json.js";
+import type { Eq } from "./Eq.js";
+import type { Ord } from "./Ord.js";
+import { Buffer } from "node:buffer";
```
bladyjoker, line 6: I don't think this will work in the browser. Keep in mind we need to keep this available to web AND node envs.

bladyjoker: Please check and comment.

runtimes/typescript/lbr-prelude/src/LambdaBuffers/Runtime/AvlTree.ts
```diff
@@ -0,0 +1,270 @@
...
+ *
+ * Recall that an AVL tree satisfies the invariants:
+ *
+ * - the absolute value of the difference of height between `left` and `right` is
+ *   at most `1`.
+ */
+export type Node<K> =
+  | null
+  | { element: K; height: number; left: Node<K>; right: Node<K> };
+
```
bladyjoker, line 14: This is killing me, we can't maintain this whole lot.

Is there no libraries we can reuse that seem stable? https://github.com/AlexAegis/avl/tree/master looks like smtn, but I would hope that there's the defacto standard containters lib widely used

bladyjoker (commented):
First round! That's a lot of work you put in! Amaz!

runtimes/typescript/lbr-prelude/src/LambdaBuffers/Runtime/Prelude.ts
```diff
@@ -307,7 +307,7 @@ export const jsonText: Json<Text> = {
 export type Maybe<A> = A | undefined;
 
 /**
- * `Eq` instance for `Maybe<A>`
+ * {@link Eq} instance for {@link Maybe}
```
bladyjoker, line 310: Why don't you do the same thing you did with Either? Would rather we stick to a pattern for representing enums rather than flerting with undefined

extras/flake-typescript.nix
```diff
@@ -0,0 +1,185 @@
...
+    '';
+
+  };
+
+  # Run tests with `npm test`.
+  test = node2nixDevelopAttrs.package.override
+    {
+      postInstall =
+        ''
+          npm --loglevel=verbose test
```
bladyjoker, line 158: What does this command actually do?

extras/flake-typescript.nix
```diff
@@ -0,0 +1,185 @@
...
+in
+{
+  devShells = {
+    # Note: when using the devshell, when in the project root directory, type
+    # ```
+    # ln -s "$NODE_PATH" node_modules
+    # ```
+    # so when `node` executes, it will use the dependencies gathered together
+    # by nix
+    "${name}" = shell;
```
bladyjoker, line 173: Let's prepend or append `typescript` to these names? I'm not sure which format we used for other lang flakes, we should probably align.

runtimes/typescript/lbr-plutus/build.nix
```diff
@@ -0,0 +1,15 @@
+{ ... }:
+{
+  perSystem = { config, ... }:
+    let
+      typescriptFlake =
+        config.overlayAttrs.extras.typescriptFlake {
+          name = "lbr-plutus-typescript";
+          src = ./.;
+          dependencies = [ config.packages."lbr-prelude-typescript-tarball" ];
```
bladyjoker, line 9: This is where we need to also specify `devShellHook` (which installs pre-commit command when in the devShell), `devShellTools` etc.

bladyjoker, line 9: Do you think it's better to specify tarballs in deps or pure source? If it's tarballs, then we need to expose the 'tarballing' function, but also maybe not as typescriptFlake makes it for you! Ignore me

runtimes/typescript/lbr-plutus/build.nix
```diff
@@ -0,0 +1,15 @@
+{ ... }:
+{
+  perSystem = { config, ... }:
+    let
+      typescriptFlake =
+        config.overlayAttrs.extras.typescriptFlake {
+          name = "lbr-plutus-typescript";
```
bladyjoker, line 7: I wonder if we should rather name this `lbr-plutus` and let typescriptFlake add the `-typescript` suffix. I think that's what we do elsewhere?

extras/flake-typescript.nix
```diff
@@ -0,0 +1,185 @@
...
+        then { echo 'No `package-lock.json` provided. Running `npm install --package-lock-only` may fix this'; exit 1; }
+        fi
+
+        #########################################################
+        # Use `npm` to add the extra tarball dependencies from nix.
+        #########################################################
+
+        # Note `npm` needs to modify these files, so we change the permissions
+        chmod +777 package.json
+        chmod +777 package-lock.json
```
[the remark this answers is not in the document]
bladyjoker: Got it!

extras/flake-typescript.nix
```diff
@@ -0,0 +1,185 @@
...
+  node2nixDevelop = node2nixExprs { extraFlags = [ "--development" ]; };
+  node2nixDevelopAttrs = ((import node2nixDevelop) { inherit nodejs pkgs; inherit (pkgs) system; });
+
+  # Build the project (runs `npm run build`), and puts the entire output in the
+  # nix store
+  project = node2nixDevelopAttrs.package.override
+    {
+      postInstall =
+        ''
+          npm run --loglevel=verbose build
```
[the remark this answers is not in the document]
bladyjoker: Nah, just asking, I didn't know that was how CTL does it. Huh, we need to figure out how to document our Nix stuff.

extras/flake-typescript.nix
bladyjoker: That's quite alright! I think it's a good API/contract to have, we just need to make it clear that's how to use this.

extras/flake-typescript.nix
```diff
@@ -0,0 +1,185 @@
...
+  # Run tests with `npm test`.
+  test = node2nixDevelopAttrs.package.override
+    {
+      postInstall =
+        ''
+          npm --loglevel=verbose test
+          rm -rf $out
+          touch $out
+        '';
+    };
```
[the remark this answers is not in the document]
bladyjoker: The `buildInputs` that I can make available to the the 'testing' derivation.

See how I did it for Purs https://github.com/mlabs-haskell/lambda-buffers/blob/05cdec80aae65372c7b73b4dcdd6ce3f8cb67580/extras/flake-purescript.nix#L35
(I would have rather have testTools but here I just use whatever was supplied as devShellTools)

For devShellTools and devShellHooks see:
See https://github.com/search?q=repo%3Amlabs-haskell%2Flambda-buffers+config.settings.shell.hook+OR+config.settings.shell.tools&type=code

runtimes/typescript/lbr-plutus/src/LambdaBuffers/Runtime/PlutusLedgerApi/V1/Credential.ts
```diff
@@ -0,0 +1,314 @@
...
+  eq: (l, r) => {
+    if (l.name === "StakingHash" && r.name === "StakingHash") {
+      return eqCredential.eq(l.fields, r.fields);
+    } else if (l.name === "StakingPtr" && r.name === "StakingPtr") {
+      return LbPrelude.eqList(LbPrelude.eqInteger).eq(l.fields, r.fields);
+    } else {
+      return false;
+    }
+  },
+  neq: (l, r) => {
```
bladyjoker, line 37: Btw why neq along with eq? why not `not eq`? Sometimes more performant?

PR #155, opened 2023-12-01
Generalizing `PrintRead` to allow type information to change the generated builtin.

jaredponn:
This PR changes 
```haskell
newtype PrintRead qvn = MkPrintRead { builtins :: Map ValueName qvn }
``` 
to
```haskell
newtype PrintRead qvn = MkPrintRead { builtins :: Ref -> Maybe qvn }
``` 
s.t. the generated `qvn` may be influenced by the types it is instantiated with.

This is important because in languages without type classes, we need to explicitly say which instance should be used. For example if we have type
```
prod Dog = Dog Int Char
```
when we generate the Eq instance, writing something like 
```
eq (Dog a0 b0)  (Dog a1 b1) = a0 `eq` a1 && b0 `eq` b1
```
won't suffice, as we must provide the type information of `Int` and `Char` to ``a0 `eq`  a1`` and ``b0 `eq` b1`` respectively. We would need to write something like
```
eq (Dog a0 b0)  (Dog a1 b1) = a0 `eqInt` a1 && b0 `eqChar` b1
```
Admittedly, I see this as a temporary fix since the builtins really should provide the types they may be instantiated with e.g. we shouldn't just have `eq` as a builtin -- we should have `eq` instantiated with types`Int`, `Char`, blah blah blah, are all builtins

### DIFF
--- a/lambda-buffers-codegen/src/LambdaBuffers/Codegen/Haskell/Print/Derive.hs
+++ b/lambda-buffers-codegen/src/LambdaBuffers/Codegen/Haskell/Print/Derive.hs
@@ -8,7 +8,6 @@ import Data.Set (Set)
 import LambdaBuffers.Codegen.Haskell.Print (MonadPrint)
 import LambdaBuffers.Codegen.Haskell.Print.LamVal (printValueE)
 import LambdaBuffers.Codegen.Haskell.Print.Syntax qualified as H
-import LambdaBuffers.Codegen.LamVal qualified as LV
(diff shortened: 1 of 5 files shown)


bladyjoker (approved):
Great! Waiting for @szg251

bladyjoker:
I think this is fairly uncontroversial change and @szg251's work in Rust codegen will not suffer for it. I'll go ahead an merge this.

PR #157, opened 2023-12-07
Improved flake-typescript.nix to put packages in `npm`s cache before installing them

jaredponn:
A minor improvement which makes it s.t. one no longer needs to manually topologically sort dependencies. This was done by using the `npm cache` command which stores all package related data.


bladyjoker (approved):
Legend!

PR #161, opened 2023-12-10
Typescript codegen

jaredponn:
I'm so sorry for the large PR. If it's easier, we can book a meeting to chat about things. :P

What was done:
- Nix ops (similar to Haskell) were implemented
- Haskell code generator for typescript code
- Extremely trivial sample projects to demonstrate "things work". [Prelude sample project](https://github.com/mlabs-haskell/lambda-buffers/tree/jared/typescript-codegen/experimental/prelude-sample-project-ts), [Plutus sample project](https://github.com/mlabs-haskell/lambda-buffers/tree/jared/typescript-codegen/experimental/plutus-sample-project-ts)
- Changed TS projects (runtime systems) to use [flake-lang.nix](https://github.com/mlabs-haskell/flake-lang.nix).
- Moved [TS prelude](https://github.com/mlabs-haskell/prelude-typescript) and [TS plutus ledger api](https://github.com/mlabs-haskell/plutus-ledger-api-typescript) to separate repos

Limitations:
- Only Haskell2010 style type classes work (probably). So
  ```haskell
  instance C (f a) where ..
  instance C (T a a a) where ..
  instance C (T A B a) where ..
  instance C (T (a b) a) where ..
  instance C a where ..
  ```
  will all most likely cause strange bugs / errors in the generated Typescript code.

- Typeclasses (probably) need to have distinct names (even if they are in different modules).

Code icks that work for now, but maybe another day I'll resolve them:
- Generated Haskell code disables all typescript checks because I was having trouble convincing TS that the generated code is exhaustive. I think if I had the complete type information of every node in [`ValueE`](https://github.com/mlabs-haskell/lambda-buffers/blob/31fb021ba1c1279a24506f502fafe01b82a85af2/lambda-buffers-codegen/src/LambdaBuffers/Codegen/LamVal.hs#L27-L55), I'd then be able to completely annotate all terms in the generated TS code, and it'd probably convince TS that the generated code is okay.

### High level idea for type classes
Given a typeclass `C`, and a type `Ty`, we assume that the following exists in the `*-runtime` library:

```typescript
// the dictionary for the class C's methods
export type C<A> = { myMethod: (arg : A) => A, .. }
 // A "global" dictionary which contains all of C's instances
export const C : CInstances = { }
// The type of the global dictionary (we start with nothing in it)
interface CInstances {
}
```

Then, when generating code for `Ty`, we must generate code like
```typescript
export type Ty<A> = ... // some reasonable type defn.
export const Ty : unique symbol = Symbol('Ty') // This creates a globally unique symbol bound to the variable `Ty`
```
See [Symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) for details on the globally unique name.

And finally, when generating the instance `C (Ty a)`, we generate code like
```typescript
// Code to "extend" the type `CInstances` to include Ty as an index mapping to value`C<Ty<A>>` 
// via declaration merging.

// the actual implementation
C[Ty] = function <A>(dictA : C<A>) {
  return {
    myMethod: (a) => { ... } 
   // implement the instance for C
   }
}
```
See [Module Augmentation in Declaration Merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) for details on declaration merging.

Finally, for end users, to use the typeclass instance, they would type
```
C[Ty](Int).myMethod( ... )
```
i.e., `C[Ty](Int)` calls the instance for the class `C` with `Ty` instantiated with the type `Int` (which we assume `Int` is a unique symbol similar how `Ty` has a unique symbol)

### Uncurried functions
I think in TS, it's a bit weird to have functions return functions e.g.
```typescript
f(a)(b)
```
is a bit strange, and it's more natural (in my opinion!) to write
```typescript
f(a,b)
```

The generated code prefers the latter i.e., the code generator uncurries functions.


experimental/plutus-sample-project-ts/build.nix
```diff
@@ -0,0 +1,31 @@
+{ inputs, ... }:
+{
+  perSystem = { config, system, ... }:
+    let
+      mySchema = config.lbf-nix.lbfPlutusTypescript {
+        name = "myschema-lb";
+        src = ./lbf;
+        files = [ "MySchema.lbf" ];
+      };
```
bladyjoker, line 9: Move this to `lbf/build.nix`, make it a proper output in `packages.lbf-plutus-sample-project-ts`.

Also name the library `name = "lbf-plutus-sample-project-ts"`

Stick to the consistent naming throughout, as we do in other places, if you notice some incosistencies please note them down.

The idea is to have `lbf` libraries named per project, not per file in that project. Format: `lbf-<project-name>-<perhaps-additional-info>

experimental/plutus-sample-project-ts/lbf/MySchema.lbf
bladyjoker, line 1: I think it makes sense to call this directory `lbf` but we use `api` everywhere else so let's stick to it. The rationale being you could have a more intricate situation in actual projects

```
api/
  lbf/MySchema.lbf
  proto/my_service.proto
  postgres/my_database.sql
```

In this sample it definitely makes sense to have only `lbf`, but for consistency with other places let's use `api`

experimental/prelude-sample-project-ts/build.nix
bladyjoker, line 1: Please apply suggestions from before in plutus counterpart

extras/lbf-nix/lbf-plutus-typescript.nix
```diff
@@ -0,0 +1,39 @@
...
+  utils = import ./utils.nix pkgs;
+
+  lbfTypescript = import ./lbf-typescript.nix { inherit pkgs lbf lbg-typescript typescriptFlake; };
+  lbfTypescriptOptsForPlutus = utils.overrideAttrs
+    {
+      imports = {
+        default = { };
+        override = libs: libs // {
+          lbf-prelude = ../../libs/lbf-prelude;
+          lbf-plutus = ../../libs/lbf-plutus;
```
bladyjoker, line 19: Use `config.packages.lbf-prelude|plutus` rather than direct paths

extras/lbf-nix/lbf-plutus-typescript.nix
```diff
@@ -0,0 +1,39 @@
...
+          config.packages.lbf-plutus-typescript
+        ];
+      };
+      classes = {
+        default = [ ];
+        override = cls: cls ++ [ "Prelude.Eq" "Prelude.Json" "Plutus.V1.PlutusData" ];
+      };
+      configs = {
+        default = [ ];
+        override = _: [ ../../lambda-buffers-codegen/data/typescript-prelude-base.json ../../lambda-buffers-codegen/data/typescript-plutus.json ];
```
bladyjoker, line 34: Use `config.packages.codegen-configs` (i think that's called) rather then direct paths

extras/lbf-nix/lbf-prelude-typescript.nix
```diff
@@ -0,0 +1,42 @@
+# Build .lbf schemas that use LB Prelude package and targets TypeScript Prelude
+# (and friends) library.
+# Warning: Essentially duplicated code from  `./lbf-prelude-haskell.nix`
+# TODO(jaredponn): instead of passing everything in, how about we just "go all
+# the way" using flake-parts and make all these scripts their own flake-parts
+# module?
```
bladyjoker, line 6: I don't see them as flakes though? what would be the output config here?

extras/lbf-nix/lbf-prelude-typescript.nix
```diff
@@ -0,0 +1,42 @@
...
+            config.packages.lbf-prelude-typescript
+          ];
+      };
+      classes = {
+        default = [ ];
+        override = cls: cls ++ [ "Prelude.Eq" "Prelude.Json" ];
+      };
+      configs = {
+        default = [ ];
+        override = cfgs: cfgs ++ [ ../../lambda-buffers-codegen/data/typescript-prelude-base.json ];
```
bladyjoker, line 36: Use `config.packages.codegen-configs` rather than direct paths

extras/lbf-nix/lbf-typescript.nix
```diff
@@ -0,0 +1,282 @@
...
+            "dependencies": {
+            }
+          }
+        '';
+    };
+
+  # TODO(jaredponn): urgh, this was manually copy pastad in by hand.. automate
+  # this more nicely later... there's a huge pain point with how this needs to
+  # hit the network to create the package-lock.json...
+  packageJsonLockTemplate = with lbfTypescriptOpts;
```
bladyjoker, line 109: I see this is sad :D Updating the TS stack then requires changing this?

extras/lbf-nix/lbf-typescript.nix
```diff
@@ -0,0 +1,282 @@
...
+                "resolved": "https://registry.npmjs.org/typescript/-/typescript-5.3.3.tgz",
+                "integrity": "sha512-pXWcraxM0uxAS+tN0AG/BF2TyqmHO014Z070UsJ+pFvYuRSq8KH8DmWpnbXe0pEPDHXZV3FcAbJkijJ5oNEnWw==",
+                "dev": true
+              }
+            }
+          }
+        '';
+    };
+
+  # TODO(jaredponn): allow this to be passed in as a parameter
```
bladyjoker, line 154: Definitely

extras/lbf-nix/lbf-typescript.nix
bladyjoker, line 1: This is super involved. Damn npm

flake.nix
```diff
@@ -3,6 +3,9 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs";
 
+    # flake-lang.nix
+    flake-lang.url = "github:mlabs-haskell/flake-lang.nix/jared/add-overlay-to-flake-typescript";
```
bladyjoker, line 7: Ugh, overlays are ugly, why did you end up using that?

lambda-buffers-codegen/data/typescript-plutus.json
```diff
@@ -0,0 +1,148 @@
+{
+  "opaquesConfig": {
+    "Plutus.V1.PlutusData": [
+      "lbr-plutus/V1.js",
+      "LbrPlutusV1",
```
bladyjoker, line 5: Why is this referencign lbr stuff? and not plutus-ledger-api-ts directly?

lambda-buffers-codegen/data/typescript-prelude-base.json
```diff
@@ -0,0 +1,70 @@
+{
+  "opaquesConfig": {
+    "Prelude.Map": [
+      "lbr-prelude",
+      "LbrPrelude",
```
bladyjoker, line 5: The same, why lbr and not typescript-prelude here?

lambda-buffers-codegen/src/LambdaBuffers/Codegen/Typescript/Print.hs
```diff
@@ -0,0 +1,252 @@
...
+
+      importsDoc = vsep $ lbImportDocs ++ tsImportDocs
+   in importsDoc
+  where
+    importQualified :: (Doc ann, Doc ann) -> Doc ann
+    importQualified (pkg, mn) = "import" <+> "*" <+> "as" <+> mn <+> "from" <+> squotes pkg
+
+{- | `collectPackageDeps lbTyImports hsTyImports classImps ruleImps valImps` collects all the package dependencies.
+ Note that LB `lbTyImports` and `ruleImps` are wired by the user (as the user decides on the package name for their schemass).
+-}
```
bladyjoker, line 245: This is supposed to be where the lbTyImports and ruleImports are collected with the help of your `packages` configuration no? How does that work?

runtimes/typescript/lbr-plutus/src/LambdaBuffers/V1/Instances.ts
bladyjoker, line 1: Why is this not in upstream libs? plutus-ts and prelude-ts?

runtimes/typescript/lbr-plutus/src/LambdaBuffers/V2/Instances.ts
bladyjoker, line 1: There's like 4 modules for something that looks like reexports? Why not just use the upstream versions? What am I missign?

bladyjoker (commented):
First round done! This is awesome!

experimental/plutus-sample-project-ts/lbf/MySchema.lbf
```diff
@@ -0,0 +1,12 @@
...
+import Plutus.V1
+import Prelude (Json)
+
+sum Branchy a
+    = Node PlutusData (Extended (Branchy a))
+    | Leaf a
+
+
+derive Json (Branchy a)
+derive PlutusData (Branchy a)
```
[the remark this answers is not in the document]
jaredponn: Thanks, good catch! I added it in!

experimental/plutus-sample-project-ts/build.nix
jaredponn: Thank you for the suggestion! I changed it to as you suggested!

experimental/plutus-sample-project-ts/lbf/MySchema.lbf
jaredponn: I didn't think of it this way! Thank you for the suggestion again! I changed it to be all under `api/`

experimental/plutus-sample-project-ts/build.nix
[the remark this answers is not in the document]
jaredponn: Thank you for the suggestion! I moved them to `docs/typescript-prelude` and `docs/typescript-plutus` respectively.

extras/lbf-nix/lbf-plutus-typescript.nix
jaredponn: I see, thank you! This makes much more sense!

extras/lbf-nix/lbf-plutus-typescript.nix
jaredponn: I see, thank you!

extras/lbf-nix/lbf-prelude-typescript.nix
jaredponn: Hm, lemme backtrack a bit with what I was thinking... might've not made any sense at all xD.

It  seems to me that we use all of these `lbf-*-*.nix` files only in https://github.com/mlabs-haskell/lambda-buffers/blob/main/extras/lbf-nix/build.nix, which importantly defines the `perSystem` `lbf-nix` option. So, I suppose I was thinking that the output config of each of these `lbf-*-*.nix` files would output a `perSystem` `lbf-nix.<attr>` which would all get merged into a single attribute set. 

I believe that would be equivalent to what already exists, and wouldn't require passing everything into the files.

I may be mistaken!

extras/lbf-nix/lbf-typescript.nix
jaredponn: Yes, you understand the (rather bleak!) situation perfectly. Happy to hear some suggestions to avoid this! 

Frankly, this is just part of the pain points I've been having with `npm` in general -- it also does the same hashing / package management thing that nix does; so these systems are a bit "at odds" with each other....

extras/lbf-nix/lbf-typescript.nix
jaredponn: I would've loved to know a nicer way to approach this!

flake.nix
jaredponn: Oh goodness, this got somewhat convoluted (somewhat related PR [over here in flake-lang.nix](https://github.com/mlabs-haskell/flake-lang.nix/pull/3)

The overlay was used here:

- https://github.com/mlabs-haskell/lambda-buffers/blob/ac244a70e74d615abefd641e124ebaa4370e72eb/extras/lbf-nix/lbf-typescript.nix#L259-L282

where

1. a .lbf file was compiled (way above in the file)
2. we create a typescript project (but it's missing the extra dependencies from nix) in variable `lbTypescriptSrc`
3. so, when we create the `typescriptFlake`, we take the typescript project in the previous step, but we overlay its `postConfigure` phase to actually install the extra dependencies from nix.

Why didn't I add the extra dependencies in 2.? I suppose I was thinking code reuse so I wouldn't have to build the folder with all the extra dependencies once in 2.; and once again in 3. (3. will always do this automatically).

Perhaps there's a better way :^). Always open to suggestions :).

lambda-buffers-codegen/src/LambdaBuffers/Codegen/Typescript/Print.hs
jaredponn: This looks like a mistake on my part -- good catch with this! I updated it to s.t. it'll actually include the packages from lbTyImports + ruleImps https://github.com/mlabs-haskell/lambda-buffers/pull/161/commits/835b1e03661d924eaa589dc138c4982c12adea0d

jaredponn:
Alright, some comments for why I didn't just refer to the upstream versions of the runtime types.

tldr: not referring to the upstream types maintains an invariant in the code generator.

I mention this a bit in [High level idea for type classes at the top of this PR](https://github.com/mlabs-haskell/lambda-buffers/pull/161#issue-2034207406)

To start, we maintain the invariant

> Let `M` be a module. Then, type `T` is defined in `M` iff value `T` (a unique symbol) is defined in `M`

Why do we want this invariant?

It lets us consistently choose which instance of a type class we want. For example, suppose we have a typeclass `ModuleWithC.C` and we further assume there will be a "global dictionary" (defined in the same module) named `ModuleWithC.C` (note the same name `C`). When we want the `instance ModuleWithC.C ModuleWithT.T`, we then may type (provided the instance is already defined)

```typescript
ModuleWithC.C[ModuleWithT.T]
```

This reasoning also applies to the code generator -- this invariant provides a nice and consistent way to find the instances corresponding to a type class and a type.

So, if the LambdaBuffers code generator assumes this invariant of all types / instances it generates, all that remains is to ensure that this invariant is also satisified by the runtime system. Which is exactly why there are so many modules re-exporting the types upstream -- to satisfy the invariant expected by the code generator.

If I did refer to the upstream type, then the invariant wouldn't be satisified, and when the codegenerator tries to refer to the _value_ corresponding to a type `T`, then the value won't exist and the code generator will generate invalid code (recall that the `prelude-typescript` and `plutus-ledger-api-typescript` libraries don't have the _value_ corresponding to types as they have a much more simplistic approach to typeclasses that disregards LambdaBuffers' module system)


Hopefully this makes a bit of sense for why it is the way it is? I'm happy to hear any suggestions though!

flake.nix
bladyjoker: No strong opinions, I trust you

lambda-buffers-codegen/src/LambdaBuffers/Codegen/Typescript/Print/Derive.hs
```diff
@@ -0,0 +1,212 @@
...
+
+-- Small translation function used to help create instance dictionaries
+lamTy2PCTy :: LamTy.Types.Ty -> Maybe PC.Ty
+lamTy2PCTy = \case
+  LamTy.Types.TyRef ref -> return $ PC.TyRefI ref
+  LamTy.Types.TyVar var -> return $ PC.TyVarI var
+  LamTy.Types.TyApp _f _args (Just tyApp) -> do
+    return $ PC.TyAppI tyApp
+  _ ->
+    -- Note(jaredponn): hopefully this never happens...
```
bladyjoker, line 43: nit :D NOTE

bladyjoker (commented):
:heart:

bladyjoker (approved):
:heart: :heart: is what I wanted to say :D

lambda-buffers-codegen/src/LambdaBuffers/Codegen/Typescript/Print/Derive.hs
jaredponn: Good catch, thanks!

jaredponn:
Alright, thanks for all the help @bladyjoker . Merging this, and will write tests soon!

PR #162, opened 2023-12-12
Remove node2nix from Typescript's Nix utilities

jaredponn:
Surprisingly, one can scan the `package-lock.json` for URLS of tarballs for dependencies, then fetch those tarballs in a nix derivation, and manually add them into `npm`'s cache. Then, `npm install` will surprisingly manage to find everything in the cache without hitting the network.

# Future work
For now, this only supports grabbing things from the npm registry, and in the future we would like to support the complete set of allowed package-specs by `npm`.

Honestly, a better way forward would probably to cook up a proxy registry to the npm registry. override `npm` s.t. it always uses the proxy registry, and dump all the nix packages there. A brief glance suggests that https://verdaccio.org/ might be useful for this :^).

### DIFF
--- a/extras/flake-typescript.nix
+++ b/extras/flake-typescript.nix
@@ -40,26 +40,29 @@ pkgs:
 , ...
 }:
 let
-  # Derivation for the result of calling the CLI tool `node2nix` with the
-  # provided `src`.
-  #
-  # Notes: 
-  #
-  #   - `node2nix` creates a nix expression in `default.nix` of type 
-  #   `{pkgs, system, nodejs} -> {args, sources, tarball, package,  shell, nodeDependencies }` 
-  node2nixExprs =
-    {
-      # Extra flags passed directly to `node2nix`
-      extraFlags ? [ ]
-    }:
-    pkgs.runCommand (name + "-node2nix") { buildInputs = [ pkgs.node2nix nodejs ]; }
+  packageLockDependencies = import ./typescript/fetch-package-lock.nix pkgs { inherit src; };
+
+  # Derivation for
+  # 1. Caching (via npm cache) all dependencies already existing in the package-lock.json
+  # 2. Caching the extra dependencies provided by nix
+  # 3. Installing the extra dependencies provided by  nix to the package-lock.json
+  # 4. npm install all of the above
+  npmProject =
+    pkgs.runCommand (name + "-npm-project") { buildInputs = [ nodejs pkgs.jq ]; }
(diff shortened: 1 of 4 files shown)


runtimes/typescript/lbr-plutus/build.nix
```diff
@@ -13,27 +13,6 @@
...
-      };
-
-      # TODO(jaredponn): for some reason, writing the more terse `inherit`
-      # seems to cause infinite recursion.
-      # ```
-      # inherit (typescriptFlake) packages checks devShells;
-      # ```
-      # So instead, we explicitly write out all the attribute names and what
-      # they are assigned to.
+      inherit (typescriptFlake) packages checks devShells;
```
jaredponn, line 16: Woo hoo, no more IFD so this works now :^)

extras/typescript/fetch-package-lock.nix
```diff
@@ -0,0 +1,53 @@
+# Given a `package-lock.json`, grabs all packages as tarballs.
```
bladyjoker, line 1: Can you state which npm versions have what you expect?

extras/flake-typescript.nix
```diff
@@ -105,86 +127,101 @@ let
...
-    name = "${node2nixDevelopAttrs.args.name}.tgz";
+    name = "${name}.tgz";
 
-    # A glance at the generated nix expressions in `node2nixDevelop` will
-    # show why the following path is where the package really is.
-    src = "${project}/lib/node_modules/${node2nixDevelopAttrs.args.packageName}";
+    src = project;
     buildPhase = ''
-      cp -r "$src" .
+      cp -r "$src/." .
```
bladyjoker, line 224: Haha yes, `cp` is difficult

runtimes/typescript/lbr-plutus/build.nix
bladyjoker: hell yes!

bladyjoker (commented):
Also heads up, you have `dependencies`, but you also need `data` (dependencies). Curious how that will work

jaredponn:
> A question though. Is this really flake-typescript? or is it flake-npm? or smtn like that? At which point is typescript invoked here I can't figure it out :D

I suppose this really is just `flake-npm` -- typescript is only called with the `build` script in the `package.json` which indeed may not be typescript at all.

Unfortunately, I ended up not going with this, so I'll close this PR.

PR #163, opened 2023-12-12
Added LamVal for neq instances

jaredponn:
Quick PR to allow lambda buffers to generate neq instances.

Note that this implements the function `/=`  by defining it as `a /= b = not (a == b)` so there certainly is more room for optimization, but this gives us correctness immediately.


lambda-buffers-codegen/src/LambdaBuffers/Codegen/LamVal/Eq.hs
```diff
@@ -95,6 +95,16 @@ eqListHelper lxs rxs tys =
...
 
+-- | Essentially, takes the 'deriveEqImpl' and sticks a not in front of it.
+deriveNeqImpl :: PC.ModuleName -> PC.TyDefs -> PC.Ty -> Either P.InternalError ValueE
+deriveNeqImpl mn tydefs =
+  deriveImpl
+    mn
+    tydefs
+    (\qsum -> LamE (\l -> LamE (\r -> notE @ (eqSum qsum @ l @ r))))
+    (\qprod -> LamE (\l -> LamE (\r -> notE @ (eqProduct qprod @ l @ r))))
+    (\qrec -> LamE (\l -> LamE (\r -> notE @ (eqRecord qrec @ l @ r))))
```
bladyjoker, line 106: Curious if this can just be `notE @ (eqSum qsum)`? and similar? Probably not as app associates on the right

bladyjoker (approved):

szg251 (approved):

lambda-buffers-codegen/src/LambdaBuffers/Codegen/LamVal/Eq.hs
jaredponn: Er, I don't think this will type check. It reads as `notE` will take as argument `eqSum qsum` (of type `a -> a -> Bool`), so we'd get a type error.

PR #164, opened 2023-12-15
lbf-nix for Rust

szg251:
This PR includes the following:
- nix scaffolding for the testsuites
- appending Rust module declarations to autogen output
- fixing problem when extra sources depend on each other
- generating Cargo.toml file for autogen output
- discovering dependency crate and module names, inject the mapping to codegen


extras/lbf-nix/lbf-rust.nix
```diff
@@ -0,0 +1,154 @@
...
+    };
+
+  # 
+  crateVersions = pkgs.writeTextFile {
+    name = "lambda-buffers-crate-versions";
+    text = ''
+      num-bigint = "0.4.4"
+      serde_json = { version = "1.0.107", features = ["arbitrary_precision"] }
+      plutus-ledger-api = { git = "https://github.com/mlabs-haskell/plutus-ledger-api-rust", features = [ "lbf", ], rev = "fb93fa590908580eb40368369bf6614d42ce9a95" }
+    '';
```
szg251, line 88: This is a temporary place for this, should we put this into the `bulid.nix` file?

extras/flake-rust.nix
```diff
@@ -7,18 +7,36 @@ let
...
-    name = "lbf-rust-build-env";
-    unpackPhase = ''
-      mkdir $out
-      cp -r $src/* $out
-      cd $out
-      ${copyExtraSources}
-      ${copyData}
-    '';
-  };
+  fullSrc =
```
bladyjoker, line 13: why not; buildEnv? as is the name of the derivation. make sense imo

bladyjoker: Also please expose this as one `package` because we can literally use this to export a non-nix Rust Cargo 'bundle' that has everything needed to build.

extras/flake-rust.nix
```diff
@@ -7,18 +7,36 @@ let
...
+        unpackPhase = ''
+          mkdir $out
+          cp -r $src/* $out
+          cd $out
+          ${copyExtraSources}
+          ${copyData}
+        '';
+      };
+
+  vendoredSrc =
```
bladyjoker, line 27: some documentation would help here

extras/flake-rust.nix
```diff
@@ -68,6 +87,8 @@ in
     doInstallCargoArtifacts = true;
   });
 
+  packages."${crateName}-rust-src" = vendoredSrc;
```
bladyjoker, line 90: Let's add some documentation for what these packages are when it's not clear.

extras/lbf-nix/lbf-rust.nix
```diff
@@ -0,0 +1,154 @@
...
+    , classes ? [ ]
+    , # Dependencies to include in the Cabal's `build-depends` stanza.
+      # examples: dependencies = [ "lbf-prelude" ]
+      dependencies ? [ ]
+    , configs ? [ ]
+    , # Name of the package and also the name of the Cabal package.
+      # Examples: name = "lbf-myproject"
+      name
+    , # Version of the package and also the version of the Cabal package.
+      # Examples: version = "0.1.0.0"
```
bladyjoker, line 31: Still Cabal mentioned here, revisit docs

extras/lbf-nix/lbf-rust.nix
```diff
@@ -0,0 +1,154 @@
...
+    lbf-build.build
+      {
+        inherit src;
+        opts = {
+          inherit files;
+          import-paths = pkgs.lib.attrsets.attrValues imports;
+          gen = lbg-rust;
+          gen-classes = classes;
+          gen-dir = "autogen";
+          gen-opts = [ "--packages=${packageSet}" ] ++ builtins.map (c: "--config=${c}") configs; # WARN(bladyjoker): If I put quotes here everything breaks.
```
bladyjoker, line 63: loving the lbf-build.build!

extras/lbf-nix/lbf-rust.nix
```diff
@@ -0,0 +1,154 @@
...
+          work-dir = ".work";
+        };
+      };
+
+  cargoTemplate = opts: with (lbfRustOpts opts);
+    pkgs.writeTextFile {
+      name = "lambda-buffers-cabal-template";
```
bladyjoker, line 70: cargo

extras/lbf-nix/lbf-rust.nix
bladyjoker: I don't understand, can you elaborate?

extras/lbf-nix/lbf-rust.nix
bladyjoker, line 1: Document what this does at file level. I need to understand the entire flow here.

extras/lbf-nix/lbf-rust.nix
bladyjoker: Why is this hardcoded? Isn't this something codegen gives you? and you just fill it in?

lambda-buffers-codegen/app/LambdaBuffers/Codegen/Cli/GenRust.hs
```diff
@@ -47,3 +53,19 @@ readRustConfig f = do
...
       exitFailure
     Just cfg -> return cfg
+
+readPackages :: FilePath -> IO RS.PkgMap
+readPackages f = do
+  fExists <- doesFileExist f
+  unless
+    fExists
+    ( do
+        logError "" $ "Provided Rust Codegen package manifest file doesn't exists: " <> f
```
bladyjoker, line 63: Remove ""?

lambda-buffers-codegen/src/LambdaBuffers/Codegen/Rust/Print.hs
```diff
@@ -120,24 +122,18 @@ printImports crates =
...
+collectPackageDeps packages lbTyImports rsTyImports traitImps ruleImps valImps =
+  Set.filter
+    (/= R.MkCrateName "crate")
+    ( Set.singleton (R.MkCrateName "std")
+        `Set.union` Set.fromList [R.crateFromLbModuleName packages $ withInfo mn | (mn, _tn) <- toList lbTyImports]
+        `Set.union` Set.fromList (mapMaybe R.qualifiedToCrate $ toList rsTyImports)
+        `Set.union` Set.fromList (mapMaybe R.qualifiedToCrate $ toList traitImps)
+        `Set.union` Set.fromList [R.crateFromLbModuleName packages $ withInfo mn | mn <- toList ruleImps]
+        `Set.union` Set.fromList (mapMaybe R.qualifiedToCrate $ toList valImps)
+    )
```
bladyjoker, line 136: First thing next round, we implement this for all other backends!

runtimes/rust/lbr-prelude-derive/build.nix
```diff
@@ -9,7 +9,7 @@
           extraSources = [
             {
               name = "lbr-prelude";
-              path = ../lbr-prelude;
+              path = config.packages.lbr-prelude-rust-src;
```
bladyjoker, line 12: nice! does rust-src contain all the 'extraSources` and `extraData`?

bladyjoker: We need to be careful, because at the moment two packages A and B use `extraSources = [C]`, that might be a problem?

testsuites/lbt-plutus/api/build.nix
```diff
@@ -19,6 +19,12 @@ _: {
         files = [ "Foo.lbf" "Foo/Bar.lbf" "Days.lbf" ];
       };
 
+      packages.lbf-plutus-golden-api-rust = config.lbf-nix.lbfPlutusRust {
+        name = "lbf-plutus-rust-golden-api";
+        src = ./.;
+        files = [ "Foo.lbf" "Foo/Bar.lbf" "Days.lbf" ];
+      };
```
bladyjoker, line 26: let's goooooo woooot

testsuites/lbt-plutus/lbt-plutus-rust/build.nix
bladyjoker, line 1: bam!

testsuites/lbt-plutus/lbt-plutus-rust/src/main.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    println!("Hey")
```
bladyjoker, line 2: Hey, wassup

extras/flake-rust.nix
```diff
@@ -7,18 +7,36 @@ let
...
+          ${copyExtraSources}
+          ${copyData}
+        '';
+      };
+
+  vendoredSrc =
+    pkgs.stdenv.mkDerivation
+      {
+        src = cleanSrc;
+        name = "${crateName}-vendored-src";
```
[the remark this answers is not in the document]
bladyjoker: how's it different from build-env?

bladyjoker (approved):
Amazing work! Happy to see this come together sooo cleanly

extras/flake-rust.nix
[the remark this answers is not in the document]
szg251: This was the result of a long experimentation. 
There's a command called `cargo vendor` which will download all deps and put them into a local directory. Crane also uses this, and I thought I might be able to inject our libraries (such as `lbr-prelude`) into this vendored directory. However I bumped into a few issues, so went back to a simpler solution. However, I still have the feeling that this could be achieved.
Now, this will only grab the source and modify the Cargo.toml such that extra-sources point to `../` instead of `.extras`

extras/lbf-nix/lbf-rust.nix
szg251: with codegen, we only get the dependency names (e.g. `num-bigint`), but not the versions
what I'm doing a few lines below is kind of a mapping with grep, so in the end, if a package does not require `num-bigint`, it won't be included.
I'm not sure where this information should be: we could inject it into the codegen, or it could be a solely nix thing

extras/flake-rust.nix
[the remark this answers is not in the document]
szg251: Also, vendored sources won't include an .extras folder, expecting that it is itself inside one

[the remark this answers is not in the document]
szg251: I'll document this

runtimes/rust/lbr-prelude-derive/build.nix
szg251: rust-src does not include other dependencies, they have to be included in the build.nix, and located in the .extras folder. This is to avoid problems with circular dependencies

testsuites/lbt-plutus/lbt-plutus-rust/src/main.rs
szg251: This will be removed in the next PR

extras/lbf-nix/lbf-rust.nix
szg251: I'm creating an issue for this, I need to think about where to put these default versions.

szg251: https://github.com/mlabs-haskell/lambda-buffers/issues/167

lambda-buffers-codegen/app/LambdaBuffers/Codegen/Cli/GenRust.hs
szg251: I think this is necessary:
https://github.com/mlabs-haskell/lambda-buffers/commit/ff570a79a0899e223e8c57070324e91631049185

PR #184, opened 2024-02-02
Use simplified rustFlake + handle lb package versions

szg251:
Depends on: https://github.com/mlabs-haskell/flake-lang.nix/pull/19

Added a new `extraVersions` to lbfRust, to manually override dependencies.
Defaults can be found here: https://github.com/mlabs-haskell/lambda-buffers/blob/e716d60a822f5ad06dad5bac8115dca07e0f0752/extras/lbf-nix/lbf-rust.nix#L85-L90
Non-matching dependencies will use `{"path": "../LIB_NAME-0.1.0" }`, where LIB_NAME is the name of the build.json dependency


bladyjoker (approved):
I dig! I do worry a bit whether we want to specify versions in Cargo.toml or somehow let the pkgset provided by the environment decide on it?

The same issue we have with Cabal, we largely stopped setting version bounds and we let the Nix provided build environment decide what's what.

Thoughts?

szg251:
> I dig! I do worry a bit whether we want to specify versions in Cargo.toml or somehow let the pkgset provided by the environment decide on it?

The problem we're having here is actually with the generated Cargo.toml file for lambda buffers rust packages. There are two cases:
- extraSources dependencies: there's only one possible version (usually), but with a version flag in the folder name (e.g. lbf-prelude-0.1.0). We (nix) has to know the exact path, including the version number.
- normal dependencies from crates.io: these versions must align with other deps, for e.g. lbr-prelude is using num-bigint 0.4.4, other versions would fail. I used more flexible versions, to lessen the burden of updates.

PR #187, opened 2024-02-03
Flake lang migration rust

jaredponn:


szg251:
I think this is superseded by https://github.com/mlabs-haskell/lambda-buffers/pull/184

PR #190, opened 2024-02-07
Attempt at reducing `nix flake lock` times

jaredponn:
`flake-lang` has like 200 ish instances of `nixpkgs` and every time we include it, that's about +200 more instances of `nixpkgs`? So, since `plutus-ledger-api-typescript` and `prelude-typescript` depend on `flake-lang`; and we include both of those in this repo, that's like an extra 400 ish instances of `nixpkgs` that we should be able to get rid of.

In this PR, 
- I updated `plutus-ledger-api-typescript` to follow `flake-lang`
- and in the [`plutus-ledger-api-typescript` repo](https://github.com/mlabs-haskell/plutus-ledger-api-typescript/), I make it follow `flake-lang` in https://github.com/mlabs-haskell/plutus-ledger-api-typescript/pull/23.

### Stats
Prior to this PR, we had:
```
$ cat flake.lock | jq '.nodes | keys' | grep nixpkgs | wc -l
718

$ time nix flake lock
real	0m45.705s
user	0m45.321s
sys	0m0.305s

```

This PR:
```
$ cat flake.lock | jq '.nodes | keys' | grep nixpkgs | wc -l
351

$ time nix flake lock
real	0m8.394s
user	0m8.300s
sys	0m0.081s


```
In raw terms of speed, this is like 5x faster, and we lost ~400 `nixpkgs` which was what we expected from the counting above.

### Other notes for `flake-lang` (CTL is the major culprit for extra nixpkgs)
flake-lang has about 200 nixpkgs, most of them come from CTL (120 of them).

So, looking at CTL, the major culprit of about 100 of CTL's nixpkgs is `db-sync`. Thus, if we remove https://github.com/Plutonomicon/cardano-transaction-lib/blob/e23ff270b3013db769c0dba1e7cfb284f4517a82/flake.nix#L76, CTL goes from having about 120 instances of nixpkgs to about 30.

I suppose if we change the `cardano-db-sync` input to the latest version + make it follow the same `haskellNix` (see below), we get CTL completely down to about 30ish instances of nixpkgs
```
    db-sync = {
        inputs.haskellNix.follows = "haskell-nix";
        url = "github:input-output-hk/cardano-db-sync/13.1.1.3";
    };
```

I'll remark that this change makes CTL's `nix flake lock` about 3x faster but it's all under 2 seconds anyways.


bladyjoker (approved):
Hell yes! thank you

PR #193, opened 2024-02-08
Draft of following flake-lang everywhere

jaredponn:


flake.nix
```diff
@@ -1,31 +1,35 @@
...
     # Hercules CI effects
-    hci-effects.url = "github:hercules-ci/hercules-ci-effects";
+    hci-effects.follows = "flake-lang/hci-effects";
 
     # Nix library for Google Protobufs
-    proto-nix.url = "github:mlabs-haskell/proto.nix";
+    proto-nix = {
+      url = "github:mlabs-haskell/proto.nix";
+      inputs.nixpkgs.follows = "nixpkgs";
+      inputs.haskell-nix.follows = "flake-lang/haskell-nix";
```
jaredponn, line 22: I suppose we could make _all_ of [proto.nix](https://github.com/mlabs-haskell/proto.nix/blob/main/flake.nix)'s inputs follow our stuff, but I don't think it's significant enough of a difference to warrant this change.

bladyjoker (approved):

bladyjoker (approved):
Thanks! If you could make sure the lb-cardano-demo works when these changes are done that would be great. It's basically our only canary test :D

jaredponn:
Okay, some final numbers

```
$ time nix flake lock
real	0m12.575s
user	0m12.342s
sys	0m0.264s

$ cat flake.lock | jq '.nodes | keys' | awk -e 'BEGIN { i = 0 } /nixpkgs/ { i++ } END {print i }'
375
```

### Further work

If we want to get this to back to be the numbers in the original PR, we'll need to modify the plutus-ledger-api-rust since that's contributing a significant overhead e.g. the following patch
```diff
diff --git a/flake.nix b/flake.nix
index 792c091..5b02a10 100644
--- a/flake.nix
+++ b/flake.nix
@@ -39,10 +39,10 @@
     };
 
     # Rust runtime
-    plutus-ledger-api-rust = {
-      url = "github:mlabs-haskell/plutus-ledger-api-rust";
-      inputs.nixpkgs.follows = "nixpkgs";
-    };
+    # plutus-ledger-api-rust = {
+    #   url = "github:mlabs-haskell/plutus-ledger-api-rust";
+    #   inputs.nixpkgs.follows = "nixpkgs";
+    # };
   };
 
   outputs = inputs@{ flake-parts, ... }:
```
gives us the following
```
$ time nix flake lock

real	0m0.942s
user	0m0.852s
sys	0m0.066s

$ cat flake.lock | jq '.nodes | keys' | awk -e 'BEGIN { i = 0 } /nixpkgs/ { i++ } END {print i }'
117
```
(of course, we wouldn't want to just comment it out because we probably need it somewhere -- I'm fairly certain the problem is the cyclic reference from lambda-buffers to plutus-ledger-api-rust so we could then make plutus-ledger-api-rust follow lambda-buffer's flake-lang)

bladyjoker (approved):
Sorry for the delay!

szg251 (approved):

PR #195, opened 2024-02-12
Recursively copy symlinks instead of copying the symlink for lbf haskell

jaredponn:
Following some slack discussion, we noticed that on some people's computers we can't build certain derivations from lbf haskell because their computer couldn't follow a symlink.

So, this PR will copy the symlink recursively instead.

### DIFF
--- a/extras/lbf-nix/lbf-haskell.nix
+++ b/extras/lbf-nix/lbf-haskell.nix
@@ -100,7 +100,7 @@ let
 
       installPhase = ''
         mkdir -p $out;
-        cp -r autogen $out
+        cp -rL autogen $out
         cp ${name}.cabal $out/${name}.cabal;
         find $out;
 


bladyjoker (approved):
Can you do a quick scan and see if other nix libs do this? Thanks!!!

jaredponn:
> Can you do a quick scan and see if other nix libs do this? Thanks!!!

Super interesting!

In other parts of lambda-buffers, we do something like
```
cp -r autogen/* <blah blah>
```

see 
    - https://github.com/mlabs-haskell/lambda-buffers/blob/790393bd5b0766a82811bdb52afc00803f624a63/extras/lbf-nix/lbf-typescript.nix#L245
    - https://github.com/mlabs-haskell/lambda-buffers/blob/790393bd5b0766a82811bdb52afc00803f624a63/extras/lbf-nix/lbf-rust.nix#L138
    - https://github.com/mlabs-haskell/lambda-buffers/blob/790393bd5b0766a82811bdb52afc00803f624a63/extras/lbf-nix/lbf-purescript.nix#L76

so we really _copy_ the files over instead of symlinking them. So, we should have no problems there.


But I did some more digging in `nixpkgs`, and I couldn't find an instance where someone _copied the symlink_ and dumped the copy in `$out`. Indeed, it's more common for people to simply just symlink what they want i.e., something like

```
# Either of the following should be fine
ln -s $(readlink autogen) $out/src
ln -s ${lbfBuilt} $out/src
```

It kinda makes me wonder -- was copying the symlink the problem and would "re-creating" the symlink have fixed it as well? I feel like these should be the same though >.<. 

If it's of any consolation, the solution in this PR which "deep copies symlinks in `$out`" does happen in `nixpkgs` so maybe there is a reason that I'm not seeing for why people don't copy symlinks to $out e.g.  https://github.com/NixOS/nixpkgs/blob/641d3aeeab9bff18b3cb6afe0e721949d1194468/pkgs/stdenv/linux/make-bootstrap-tools.nix#L87

PR #201, opened 2024-02-23
Fix bug with lbfPlutusRust not generating Json and Eq instances

szg251:

PR #202, opened 2024-02-24
Include the user provided config for Rust and TS plutus code generator

jaredponn:
Every other code generator lets you do this, so Rust and TS should allow this for consistency with the other code generators.


szg251:
@jaredponn I think you were reading my mind
have you seen my PR? https://github.com/mlabs-haskell/lambda-buffers/pull/201

szg251:
I think we can close this (see my comment above)

PR #216, opened 2024-04-09
Allow empty records for JSON generation

jaredponn:


szg251 (approved):

bladyjoker (approved):
```
record MyEmptyRecord = { }
```

Is not something LB can swallow at the moment. Why? I don't recall exactly, but it was a decision that spanned parts from the Compiler to the Codegen. Nowadays, I think we error on empty records, and we have no syntax for empty products.

However, I don't think this change is problematic. If anything, when we introduce empty product/records this would then just work.

PR #222, opened 2024-05-29
Draft of milestone 3 report

jaredponn:
Read the document here

https://github.com/mlabs-haskell/lambda-buffers/blob/jared/milestone-3-report/docs/aiken-integration.md


docs/aiken-integration.md
```diff
@@ -0,0 +1,516 @@
...
+      ]
+    )
+    (force (builtin headList))
+  ]
+)
+```
+
+Note that this would most likely offer a poor user experience as this would essentially replace a large part of Aiken's language construct with our own generated functions for constructing, deconstructing, serialization / deserialization to `Data`, etc.
+
+### Aiken's packages only support fetching dependencies remotely
```
t4ccer, line 424: Worth mentioning that there's a workaround. You can prepare build directory in a way to inject dependencies and trick aiken into thinking that they were fetched (see https://github.com/mlabs-haskell/uplc-benchmark/blob/master/nix/aiken/lib.nix#L83) but that's a very much an implementation detail

docs/aiken-integration.md
```diff
@@ -0,0 +1,516 @@
...
+      ]
+```
+
+This is awkward for LambdaBuffers since when Aiken works with the `MyRecord` type,
+it is represented with the `Constr` tag meaning that LambdaBuffers' efficient encoding would be translated to Aiken's inefficient encoding.
+Ideally, one would want to change how Aiken encodes its data types internally so that Aiken can use LambdaBuffers' efficient encoding everywhere.
+Thus, we lose all benefits of LambdaBuffers' efficient encoding when working with Aiken's mechanisms to define types because LambdaBuffers is forced to take an extra step to translate to Aiken's inefficient encoding.
+As such, Aiken's opinionated way of encoding its data is at odds with LambdaBuffers.
+
+To resolve the mismatch in the encoding of data between the two, one could alternatively sidestep all of Aiken's methods for defining types and instead use Aiken's opaque types to alias `Data` and provide ones own constructors / record accesses as follows
```
t4ccer, line 338: You could also define record as usual `pub type MyRecord {data: Int}` and write a function `Data -> MyRecord`, and take datum and redeemer as `Data`, so you maintain dot record access and you do decoding only once (but not longer on demand) though you'd have to benchmark that

docs/aiken-integration.md
```diff
@@ -0,0 +1,516 @@
...
+If LambdaBuffers compiled `MyRecord` to a [record in Aiken](https://aiken-lang.org/language-tour/custom-types) as follows
+
+```rust
+type MyRecord {
+    a : Int,
+    b : Int
+}
+```
+
+we know that Aiken will internally represent this as the following `Data` type
```
t4ccer, line 186: Emphasizing on "we know" being "we've observed" because docs don't mention them and push you to use the json schema thingie

docs/aiken-integration.md
```diff
@@ -0,0 +1,516 @@
...
+For example, the following Aiken type
+
+```ruost
+type EqDict<a> {
+    eq: fn(a,a) -> Bool, 
+    neq: fn(a,a) -> Bool, 
+}
+```
+
+would produce an error like
```
t4ccer, line 38: Add somewhere which version of Aiken you've used please. it gives nice error message only on master, stable release crashes (as of v26)

docs/aiken-integration.md
jaredponn: I'm not quite sure I follow what we're saying here.

Are we saying:

We define
```
pub type MyRecord { data : Int }
```

then we define a function

```
fn fromData(theData : Data) -> MyRecord { ... }
```

and we write validators like
```
validator {
    pub fn hello_world(redeemer: Data, _scriptContext: Data) {
        let theRecord = fromData(redeemer)

        ...

    }
}
```

If so, this is what the text suggests in the example [here](https://github.com/mlabs-haskell/lambda-buffers/blob/jared/milestone-3-report/docs/aiken-integration.md?plain=1#L72-L146). I suppose this wasn't clear -- were we suggesting to include a sentence saying: one can encode and decode `Data` for validators in Aiken explicitly with LB's translation once at the beginning of a validator?

> you maintain dot record access and you do decoding only once

I think this is what we want avoid -- I don't see strong reasoning for using LB's efficient encoding if Aiken doesn't use it.

borja-bonet-segui (approved):

jaredponn:
Alright, I think this should be mostly in shape for sending it off (fingers crossed!).

To see the milestone report on the website, type
```
nix develop .#dev-docs
cd docs
mdbook build
chromium book/index.html
```

borja-bonet-segui (approved):

PR #229, opened 2024-06-17
Move all lambda buffers related code from PLA to LBF

szg251:
We've been discussing this for a while, we can now remove lbf dependencies in plutus-ledger-api-rust repo, making the flake evaluation lighter (removing circular dependency). This also makes sense from an architectural standpoint too.

### Drawbacks:

- unable to separate `lbr-prelude` and `lbr-plutus` because of the Rust orphan rule, the`Json` trait and it's implementations must be in the same package
- cannot use the derive macro for Json, so we had to manually implement Json instances (I pasted in and refactored the derived code)


runtimes/rust/lbr-prelude/src/plutus/lamval.rs
bladyjoker, line 1: This can't be here, this is lbr-plutus, please separate these things as inconsistencies here are very hard to keep track of

bladyjoker (commented):
I'm not at all convinced about this unfortunately. Having Plutus related stuff in lbr-prelude feels quite wrong.

My 2 cents is that the previous setup felt much cleaner and responsibilities were placed more appropriately.

runtimes/rust/lbr-prelude/src/plutus/lamval.rs
szg251: I think I remembered that you also wanted this..
we cannot separate the `Json` trait from the implementations, so it has to be in lbr-prelude. To mitigate the issue, I introduced a feature flag for this, so users who don't need plutus can skip this part.

szg251: https://mlabs-corp.slack.com/archives/C0441AEUJTZ/p1707728195606979?thread_ts=1707655526.821769&cid=C0441AEUJTZ

Issue #237, opened 2024-07-10
Rust codegen Phantom handling bug


bladyjoker:
Phantom handling in Rust codegen is not correct I believe https://github.com/mlabs-haskell/lambda-buffers/blob/401f8a920a557c71440795174da199a1e128c4f9/lambda-buffers-codegen/src/LambdaBuffers/Codegen/Rust/Print/TyDef.hs#L177

```
record CreateScriptsRequest a = {
  scripts : Map ScriptHash (ScriptDatum a)
  }
derive Eq (CreateScriptsRequest a)
derive Json (CreateScriptsRequest a)

record ScriptDatum a = {
  scriptType : a,
  -- ^ Protocol specific script type stored in this UTxO
  redeemer : AssetClass
  -- ^ Wallet containing `$redeemer` must sign the 'delete-script' transaction as an authorization method
}
derive Eq (ScriptDatum a)
derive PlutusData (ScriptDatum a)
derive Json (ScriptDatum a)

```
creates a

```rust
#[derive(std::fmt::Debug, std::clone::Clone)]
pub struct CreateScriptsRequest<A>{pub scripts: lbf_prelude::prelude::Map<lbf_plutus::plutus::v1::ScriptHash
                                  ,crate::cardano::extra::scriptstorage::validation::ScriptDatum<A>>,
                                  phantom_A: std::marker::PhantomData<A>}

#[derive(std::fmt::Debug, std::clone::Clone)]
pub struct ScriptDatum<A>{pub script_type: A,
                         pub redeemer: lbf_plutus::plutus::v1::AssetClass}
```

There's 2 problems here.
1. There's no actual phantom types in the original LB type def
2. Even if there was, we have a problem when phantom values are not made pub in Rust

bladyjoker:
In #239 we solved the issue with phantomdata fields not being public. We still have the issue with phantoms analysis.

PR #239, opened 2024-07-10
Rust's phantom handling fixes

bladyjoker:
DONE
 - [x] Record TyDef printing adds `pub` to phantom record fields
 - [ ] Phantom analysis bug fix is still unresolved (https://github.com/mlabs-haskell/lambda-buffers/issues/237)


szg251 (approved):
looks good

PR #247, opened 2024-09-10
Milestone 4 Report

jaredponn:
This PR:
- [x] Milestone 4 report to demonstrate proof of completion


szg251 (approved):

bladyjoker (changes requested):
This looks great! Thanks a lot!

One thing I'm missing is wrt versioning is something like what @szg251 did in https://github.com/mlabs-haskell/tx-village/blob/main/CONTRIBUTING.md

@szg251 can you advise on the scheme for LB? We need to version lbf schemas, runtimes and the frontend/compiler/codegen.

bladyjoker (approved):
Thanks mate!

PR #249, opened 2024-09-16
Refactoring logging functions / make the error messages more closely match `gcc`s output

jaredponn:
This PR:
- [x] Refactored all logging (`logError` and `logInfo` functions (from `lambda-buffers-frontend` and `lambda-buffers-codegen`) to reuse the logging functions in `lambda-buffers-compiler` since `lambda-buffers-compiler` is an existing dependency of  both `lambda-buffers-frontend` and `lambda-buffers-codegen`
- [x] Change error / info messages to be dumped to `stderr` instead of `stdout` (this is more aligned with other compilers' behaviors)
- [x] Shuffled error messages to more closely follow what `gcc` outputs e.g. 
  ```
  $ lbf-prelude-to-haskell DoesNotExist.lbf
  DoesNotExist.lbf: error: the provided module file DoesNotExist.lbf doesn't exist
  
  $ cat TEST.lbf
  module TEST
  
  import Prelude (Eq, Json, Maybe, Either, List, Integer)
  
  derive Eq What
  $ lbf-prelude-to-haskell TEST.lbf
  TEST.lbf:5.11-5.15: error: [module TEST] type What not found in the module's scope Either Integer List Maybe Prelude.Either Prelude.Integer Prelude.List Prelude.Maybe
  
  $ cat TEST.lbf
  module TEST
  
  import Prelude (Eq, Json, Maybe, Either, List, Integer)
  
  prod What = (Integer
  derive Eq What
  
  $ lbf-prelude-to-haskell TEST.lbf
  TEST.lbf:6.7: error:
  unexpected keyword
  expecting s-expression
  ```


lambda-buffers-compiler/src/LambdaBuffers/Logger.hs
```diff
@@ -0,0 +1,41 @@
...
+
+import System.IO qualified
+
+{- | Logs an INFO message of the format
+> <file>: info: <msg>
+if @<file>@ is non empty, otherwise, it prints
+> info: <msg>
+-}
+logInfo :: FilePath -> String -> IO ()
+logInfo "" msg = System.IO.hPutStrLn System.IO.stderr $ "info: " <> msg
```
szg251, line 18: Shouldn't this be `System.IO.stdout` for an info log?

bladyjoker: Hmm not sure what I was thinking here, but I vaguely remember using stderr for any logging for some reason

bladyjoker (approved):
This looks great!

I left a comment about `extras/lambda-buffers-utils` to keep the compilation level separation. Let me know if that works

lambda-buffers-frontend/lambda-buffers-frontend.cabal
```diff
@@ -123,6 +123,7 @@ executable lbf
     , directory
     , filepath                    >=1.4
     , lambda-buffers-codegen-pb   >=0.1
+    , lambda-buffers-compiler     >=0.1
```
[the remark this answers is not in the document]
jaredponn: This sounds good! I put it in `extras/lambda-buffers-utils`!

extras/lambda-buffers-utils/cabal.project
```diff
@@ -0,0 +1,6 @@
+packages: ./.
+
+tests: true
+
+-- TODO(bladyjoker): Remove once freer-simple fixes the template-haskell bounds
+allow-newer: freer-simple:template-haskell
```
bladyjoker, line 6: you can probably delete this?

extras/lambda-buffers-utils/src/LambdaBuffers/Logger.hs
```diff
@@ -0,0 +1,41 @@
+{- | Module: LambdaBuffers.Logger
+
+Utilities for logging INFO, and ERROR messages.
+
+This follows the GNU error message format:
+<https://www.gnu.org/prep/standards/html_node/Errors.html>
+-}
+module LambdaBuffers.Logger (logInfo, logError, logErrorWithSourceSpan) where
```
bladyjoker, line 8: Should we put this under LambdaBuffers.Utils? As you prefer

bladyjoker (approved):
Thank youuuu!!!

extras/lambda-buffers-utils/.envrc
```diff
@@ -0,0 +1 @@
+use flake ..#dev-utils
```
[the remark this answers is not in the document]
jaredponn: Should be done in https://github.com/mlabs-haskell/lambda-buffers/pull/249/commits/d325b7898a0374fdb861badbda4852d46bb9d1fe


Note that the other devshells were called e.g. `dev-codegen`, and I also refactored that to `dev-lambda-buffers-codegen` s.t. it's all consistent

extras/lambda-buffers-utils/src/LambdaBuffers/Utils/Logger.hs
```diff
@@ -0,0 +1,41 @@
...
+
+import System.IO qualified
+
+{- | Logs an INFO message of the format
+> <file>: info: <msg>
+if @<file>@ is non empty, otherwise, it prints
+> info: <msg>
+-}
+logInfo :: FilePath -> String -> IO ()
+logInfo "" msg = System.IO.hPutStrLn System.IO.stderr $ "info: " <> msg
```
szg251, line 18: shouldn't this be `stdout`?

szg251 (commented):
I thought I reviewed this earlier...

extras/lambda-buffers-utils/src/LambdaBuffers/Utils/Logger.hs
jaredponn: Honestly, I don't think so -- most of the time `stdout` is reserved for the actual output of the program, and `stderr` is for printing diagnostic and error messages. The actual output of compilers is usually done on just files, and everything else is then just diagnostic / errors which hence should go on stderr.

In the case of LB, `info` messages generally seem like "diagnostic" and hence **should** go to `stderr`.

But I'll change it to `stdout` anyways :P.

References
- https://linux.die.net/man/3/stderr
  where we quote
  > Under normal circumstances every UNIX program has three streams opened for it when it starts up, one for input, one for output, and **one for printing diagnostic or error messages**.

jaredponn:
I'll merge this since it seems we've address all PR feedback

PR #250, opened 2024-09-24
Added a `CHANGELOG.md` and `CONTRIBUTING.md` documentation

jaredponn:
This PR:
- [x] added an empty `CHANGELOG.md` for all of LB
- [x] added a `CONTRIBUTING.md` which is an abridged version of tx-village's change log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Initial release
-
 ### Schemas
 
 #### lbf-prelude
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,18 +7,81 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
+- Initial release
 
-### Changed
+### Schemas
 
-### Removed
+#### lbf-prelude
+##### Purescript
+##### Rust
+##### Haskell
+##### Typescript
 
-## v1.0.0
+#### lbf-plutus
+##### Purescript
+##### Rust
+##### Haskell
+##### Typescript
+##### Plutarch
+##### PlutusTx
 
-### Added
+### Tools
 
-- Initial versioning release
+#### Frontend
+#### Compiler
+#### Compiler Proto API
+#### Codegen
+#### Codegen Proto API
 
-### Changed
+### Runtimes
 
-### Removed
+#### Rust lbr-prelude
+#### Haskell lbr-prelude
+#### Typescript lbr-prelude
+
+#### Rust lbr-plutus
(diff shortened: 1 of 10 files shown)


bladyjoker (approved):
This is great and am certain probably sufficient.

I'd like to have the following somehow done;

```markdown
# LB v.1.0.0

## Schemas

lbf-prelude v.1.0.0
  - Purescript
     - <lib.ver> of the libraries hosting the types and implementations (basically from the codegen config)  
   - Rust
   - Haskell
   - Typescript

lbf-plutus v.1.0.0 
  - Purescript
     - <lib.ver> of the libraries hosting the types and implementations (basically from the codegen config)  
   - Rust
   - Haskell
   - Typescript
   - Plutarch
   - PlutusTx

## Tools

Frontend v.1.0.0
Compiler v.1.0.0
   - Proto API v.1.0.0
Codegen v.1.0.0
   - Proto API v.1.0.0

## Runtimes

Rust lbr-prelude v.1.0.0
Haskell lbr-prelude v.1.0.0
Typescript lbr-prelude v.1.0.0
 
Rust lbr-plutus v.1.0.0
Haskell lbr-plutus v.1.0.0
PlutusTx lbr-plutus v.1.0.0
Plutarch lbr-plutus v.1.0.0
Typescript lbr-plutus v.1.0.0
```

Basically our repo wide version and how does it reflect the underlying versions of schemas, the libraries they map to, runtimes, tools and their external APIs.

I suspect we should writeup how the versioning numbers change and stay aligned (it at all) between all this parts.

And also we should update the Cabal,Dhall,Cargo,Proto,LB build configurations and specify the versions explicitly if not already so.

I don't have the cycles to be more precise about this tbh, so whatever you can do here it's good enough!

szg251:
I think we need to adjust the versioning, as PVP is only appliable to Haskell (and PureScript I guess).
My strategy for polyglot repos is to set the first number of the PVP version to 1 all times, so the versions can be consistent:
e.g.: semver 7.8.0 -> PVP 1.7.8.0
with the exception of preproduction versions, which can start with a 0
See the contribution guideline here https://github.com/mlabs-haskell/plutus-ledger-api-rust/blob/main/CONTRIBUTING.md, and oc feel free to propose changes.

jaredponn:
@szg251 right, I see what you're saying now about the prefixing all PVP with 1 or 0.

I copied exactly what was written for your versioning scheme with the following changes:

- I didn't enforce an alignment between versions of the entire repo and its subparts. It seems to me that forcing an alignment of major version numbers would contradict the fact that we bump the major version numbers only if there is a breaking change.
- Instead of having a CHANGELOG.md for each subproject, we have a CHANGELOG.md for everything at the root of the project. I dunno what's better, but we'll see with time what works out :)

jaredponn:
I'm going to merge this s.t. we can get the Milestone 4 write up finalized, and if we have any other changes we'd like (I'm sure we'll find some!) we can put it in a separate PR :P.

PR #252, opened 2024-10-09
Added the targeted libraries the lbf schemas target (when applicable)

jaredponn:
This PR:
- [x] Documentation for which library the lbf schema targets in the `CHANGELOG.md`
- [x] Add the PureScript runtimes to the `CHANGELOG.md` (this was missing previously, and start with major version 1 to be more consistent with all other packages)


jaredponn:
> Should we add the same for lbf-prelude or?

Good suggestion! I added this in as well :)

PR #298, opened 2026-08-14
DO NOT MERGE THIS 

gnumonik:
This gets the checks passing for the haskell backends after adding BLS primitives to the LB prelude, but this is almost certainly not the _right_ way to do things (which requires putting Koz's package on hackage). 

I'm putting this up primarily so I can look at a pretty GH diff and more easily keep track of what I've changed.


libs/lbf-prelude/Prelude.lbf
bladyjoker, line 1: Rather put this in libs/lbf-plutus/Plutus/V4.lbf?

